PR #164, opened 2015-10-20
allow-delete

TakenPilot:
New Feature
- Allow DELETE on `/uris/` and `/pages`

### DIFF
--- a/lib/routes/uris.js
+++ b/lib/routes/uris.js
@@ -7,7 +7,7 @@
 'use strict';
 
 var responses = require('../responses'),
-  db = require('../services/db');
+  controller = require('../services/uris');
 
 /**
  * @param req
@@ -15,7 +15,7 @@ var responses = require('../responses'),
  */
 function getUriFromReference(req, res) {
   responses.expectText(function () {
-    return db.get(req.uri);
+    return controller.get(req.uri);
   }, res);
 }
 
@@ -25,11 +25,17 @@ function getUriFromReference(req, res) {
  */
 function putUriFromReference(req, res) {
   responses.expectText(function () {
-    if (req.uri === req.body) {
-      throw new Error('Client: Cannot point uri at itself');
-    }
+    return controller.put(req.uri, req.body);
+  }, res);
+}
 
-    return db.put(req.uri, req.body).return(req.body);
+/**
+ * @param req
+ * @param res
+ */
+function deleteUriFromReference(req, res) {
+  responses.expectText(function () {
+    return controller.del(req.uri);
   }, res);
 }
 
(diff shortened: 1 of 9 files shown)


lib/routes/uris.js
```diff
@@ -39,11 +45,13 @@ function routes(router) {
...
+  router.all('/:name', responses.methodNotAllowed({allow: ['get', 'put', 'delete']}));
   router.all('/:name', responses.notAcceptable({accept: ['text/plain']}));
   router.get('/:name', getUriFromReference);
   router.put('/:name', putUriFromReference);
-  router.post('/', responses.notImplemented);
+  router.delete('/:name', deleteUriFromReference);
+
 }
 
 module.exports = routes;
```
TakenPilot: Moved business/blocking logic to controller.

test/api/pages/delete.js
```diff
@@ -0,0 +1,68 @@
...
+
+      acceptsJson(path, {name: 'valid', version: 'valid'}, 200, pageData);
+      acceptsJson(path, {name: 'valid', version: 'missing'}, 404, { message: 'Not Found', code: 404 });
+
+      acceptsJsonBody(path, {name: 'valid', version: 'valid'}, pageData, 200, pageData);
+      acceptsJsonBody(path, {name: 'valid', version: 'missing'}, pageData, 404, { message: 'Not Found', code: 404 });
+
+      acceptsHtml(path, {name: 'valid', version: 'valid'}, 406, '406 text/html not acceptable');
+      acceptsHtml(path, {name: 'valid', version: 'missing'}, 406, '406 text/html not acceptable');
+    });
```
cruzanmo: Should there be a test confirming that delete is not allowed with :name.:ext?

TakenPilot: Good idea!  It should at least be testing the error message.

test/api/pages/delete.js
```diff
@@ -0,0 +1,68 @@
...
+      acceptsJson(path, {}, 405, { allow:['get', 'post'], code: 405, message: 'Method DELETE not allowed' });
+      acceptsJsonBody(path, {}, {}, 405, { allow:['get', 'post'], code: 405, message: 'Method DELETE not allowed' });
+      acceptsHtml(path, {}, 405, '405 Method DELETE not allowed');
+    });
+
+    describe('/pages/:name', function () {
+      var path = this.title;
+
+      acceptsJson(path, {name: 'valid'}, 200, pageData);
+      acceptsJson(path, {name: 'missing'}, 404, { message: 'Not Found', code: 404 });
```
cruzanmo: Per slack conversation, there some debate over whether a DELETE should return a 404, since DELETE can be [considered idempotent](https://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol#Idempotent_methods_and_web_applications)

Since we are returning the deleted object (so that the client has the most recent version of what was deleted), then 404 makes sense to me. Essentially, we are forcing a GET prior to DELETE.

TakenPilot: Pretty much!  : )

PR #166, opened 2015-10-27
allow .yml when looking for schema

### LINKED ISSUE #165: schema route not picking up schema.yml
It works for `schema.yaml` in components, but not `schema.yml`

yoshokatana:
fixes #165

### DIFF
--- a/lib/bootstrap.js
+++ b/lib/bootstrap.js
@@ -9,6 +9,8 @@ var path = require('path'),
   log = require('./log'),
   chalk = require('chalk');
 
+_.mixin(require('lodash-ny-util'));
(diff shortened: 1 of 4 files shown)


lib/bootstrap.js
```diff
@@ -9,6 +9,8 @@ var path = require('path'),
   log = require('./log'),
   chalk = require('chalk');
 
+_.mixin(require('lodash-ny-util'));
```
yoshokatana: this was in schema.js (and wasn't used there), but is needed for bootstrapping

TakenPilot:
+1

PR #167, opened 2015-10-27
throw 404 instead of returning null when no schema

yoshokatana:
it was returning 200 + 'null' rather than returning 404 when no schema was found.


TakenPilot:
+1 when a test is added.

TakenPilot:
+1

PR #168, opened 2015-10-27
stop-double-bootstrap

TakenPilot:
Patch:
- Stop pushing to @published for all things in bootstrap
- Block writing @published or @latest to `/uris`.  All things in uris are assumed to be @published ("public") anyway, unless given a different version.


yoshokatana:
:+1:

PR #174, opened 2015-11-12
Component locals

TakenPilot:
New Feature
- PUT, POST, and DELETE now have access to locals as an additional param, which is: 
```js
const locals = {};
_.assign(locals, req.params, req.query);
locals.site = site;  //site information, like host, path, assets, etc.
```

### DIFF
--- a/lib/services/components.js
+++ b/lib/services/components.js
@@ -7,6 +7,7 @@
 'use strict';
 
 var _ = require('lodash'),
+  control = require('../control'),
   db = require('./db'),
   uid = require('../uid'),
   files = require('../files'),
@@ -34,8 +35,8 @@ function getName(uri) {
 }
 
 /**
- * @param uri
- * @param locals
+ * @param {string} uri
+ * @param {object} [locals]
  */
 function get(uri, locals) {
   var promise,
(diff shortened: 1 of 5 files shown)


lib/services/components.js
```diff
@@ -140,16 +144,17 @@ function put(uri, data) {
...
  *
  * @param {object} obj
+ * @returns {object}
  */
 function clearOwnProperties(obj) {
   _.forOwn(obj, function (value, key) {
     delete obj[key];
   });
+
+  return obj;
```
cruzanmo: Why does obj need to be returned if it is in-place?

TakenPilot: For convenience mostly.  For example, in coffee-script, the result of the last line is returned always because of this reason.

It's a useful convention to not have functions that don't return anything so that statements can be combined.

cruzanmo:
Should there be tests locals being frozen in the cascading PUT, and for PUT with locals for a single component?

TakenPilot:
That's a good idea, though the frozen interface isn't so clear about whether it should throw an exception or just ignore the write.  Different browsers do different things, for example.

What would a test like that look like?  It should fail if the contents are _not_ read-only, but would it be an exception, or just an ignored write?

Might be low ROI if the test gets too complex, maybe.  Hmm.

cruzanmo:
Maybe just check if Object.isFrozen()?

TakenPilot:
That would work.

cruzanmo:
Nice +1

PR #175, opened 2015-11-12
locals from page, reverse priority of sites index.js

TakenPilot:
New Features
- The variables `locals` is now available on a publish action from a page.
- Instances of sites can now override or add routes in front of our own routes.


yoshokatana:
:+1:

PR #177, opened 2015-11-20
pass-task

TakenPilot:
Patch
- Report what was rendered and how long it took using `hrtime`.
- Pass composer-level route with `next()` if NotFoundError

### DIFF
--- a/lib/composer.js
+++ b/lib/composer.js
@@ -365,19 +365,28 @@ function getExpressRoutePrefix(site) {
  * @param {function} next
  */
 function renderExpressRoute(req, res, next) {
-  var site, prefix, pageReference;
+  var site, prefix, pageReference, hrStart;
 
+  hrStart = process.hrtime()[1];
   site = res.locals.site;
   prefix = getExpressRoutePrefix(site) + '/uris/';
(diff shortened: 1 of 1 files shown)


lib/composer.js
```diff
@@ -365,19 +365,28 @@ function getExpressRoutePrefix(site) {
...
-      res.send(html);
-    }).catch(function (err) {
+  renderUri(pageReference, res).then(function (html) {
+    res.type('html');
+    res.send(html);
+
+    log.info('rendered express route:' +
+      (req.hostname + req.baseUrl + req.path) + ' ' +
+      ((process.hrtime()[1] - hrStart) / 1000000) + 'ms'
+    );
```
cruzanmo: Should it always log or only when there is an error?

TakenPilot: Always.  Seems we need a bigger way to hit people when they're putting logic in the GET.

TakenPilot: Right now, we're sitting at 100ms on local.  :( :( :(

cruzanmo:
+1

yoshokatana:
:+1:

PR #178, opened 2015-11-23
schedule-double

TakenPilot:
Patch:
- Change `/schedule` list to use `{_ref: }` format
- Add `@scheduled` to pages and components


yoshokatana:
:+1:  update the docs though!

yoshokatana:
:+1: :+1: :+1:

PR #179, opened 2015-11-24
resolve-media-map

TakenPilot:
New Feature
- Allow sites to modify the media map used within a template.

For example
```js
module.exports = function (router, composer) {

  router.get('/', composer);
  router.get('/:year/:month/:name', composer);

  return router;
};

module.exports.resolveMedia = function (media, locals) {
  const assetPath = locals.site.assetPath;

  if (locals.edit) {
    media.scripts = [assetPath + '/js/edit-before.js'];
  } else {
    media.scripts.unshift(assetPath + '/js/view-before.js');
    media.scripts.push(assetPath + '/js/view-after.js');
  }
};
```

### DIFF
--- a/.istanbul.yml
+++ b/.istanbul.yml
@@ -18,10 +18,10 @@ reporting:
         - lcov
     dir: ./coverage
     watermarks:
-        statements: [80, 90]
-        lines: [80, 90]
-        functions: [80, 90]
-        branches: [50, 80]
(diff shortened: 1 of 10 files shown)


TakenPilot:
I'll be putting coverage at a pure 100% this PR as well, I guess.  It's time, after all.

techmsi:
+1

PR #181, opened 2015-11-30
page-level-url

TakenPilot:
New Feature
- `/pages` now have a `url` property that must be a valid url and will not be expanded, cloned or resolved.
- On publish, a `/uris/<new page>` will be created from the url.
- Publishing will 400 if the url is missing.


yoshokatana:
:+1:

cruzanmo:
+1

PR #183, opened 2015-12-01
bootstrap-page-url

TakenPilot:
Patch
- Bootstrap will treat `url` in a page as a canonical url

Note:  We will need to say the port in config.yaml or local.yaml now for this to work.  :/


yoshokatana:
:+1:

PR #186, opened 2015-12-02
don't rely on global modules

yoshokatana:
* added `jscs`, `jshint`, and `istanbul` to `devDependencies`
* `npm test` references them rather than global versions

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -3,12 +3,6 @@ machine:
     America/New_York
   node:
     version: 4.0
-dependencies:
-  pre:
-    - npm install -g istanbul
-    - npm install -g node-gyp
-    - npm install -g jscs
-    - npm install -g jshint
 test:
   post:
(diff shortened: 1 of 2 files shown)


TakenPilot:
Can you reduce the version numbers to 
```
+    "istanbul": "^0.4",
+    "jscs": "^2.6",
+    "jshint": "^2.8",
```

Updating these things can get annoying fast, and they change often.

yoshokatana:
Yep, done.

TakenPilot:
+1 if circle works after a cache-clear.

yoshokatana:
@TakenPilot all tests pass when `node-gyp` is removed from the global deps on circle. :+1:?

yoshokatana:
All tests pass when rebuilding without cache on circle (after removing `node-gyp` from global dependencies).

PR #188, opened 2015-12-02
Notifications

TakenPilot:
New Feature
- Sites can now set webhooks for 'publish' and 'unpublish' events.

```yaml
host: 'nymag.com'
#...config, etc
notify:
  webhooks:
    published:
      - http://some-url.com/some-path
    unpublished:
      - http://some-url.com/some-path
```

The webhook will be a POST and will set a `X-Event` header with the name of the event, and will include the data from the event.

The data that the events should send has not been made clear yet, so currently is just whatever is available.  Expect that to change later.

### DIFF
--- a/lib/services/notifications.js
+++ b/lib/services/notifications.js
@@ -1,0 +1,65 @@
+'use strict';
+
+const _ = require('lodash'),
+  log = require('../log'),
+  restler = require('restler'),
+  defaultWebhook = {method: 'POST', headers: {'Content-type': 'application/json'}};
+
+/**
+ * @param {string} url
+ * @param {object} [options]
+ * @param {string} [options.method]
+ * @param {*} [options.body]
+ * @param {object} [options.headers]
+ */
+function callWebhook(url, options) {
+  restler.request(url, options).on('complete', function (result, response) {
+    log.info('called webhook', _.pick(response, ['status', 'statusText']), result);
+  });
+}
+
+/**
+ * @param {object} webhook
+ * @returns {function}
+ */
+function getNotifyWebhookCallback(webhook) {
+  const regex = new RegExp(webhook.match);
+
+  return function (op) {
+    if (op.type === webhook.op && op.key.test(regex)) {
+      if (_.isPlainObject(op.value)) {
+        let opClone = _.omit(op, 'value');
+        opClone.value = JSON.stringify(op.value);
+        op = opClone;
+      }
+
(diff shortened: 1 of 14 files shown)


test/fixtures/api-accepts.js
```diff
@@ -416,7 +417,7 @@ function beforeEachTest(options) {
   stubSchema(options.sandbox);
   stubGetTemplate(options.sandbox);
   stubMultiplexRender(options.sandbox);
-  stubLogging(options.sandbox);
+  //stubLogging(options.sandbox);
```
yoshokatana: is this supposed to be commented out?

yoshokatana:
:+1:

lib/services/notifications.js
```diff
@@ -0,0 +1,54 @@
...
+
+  if (_.isPlainObject(data)) {
+    headers[contentType] = 'application/json';
+    options.body = JSON.stringify(data);
+  } else if (_.isString(data)) {
+    headers[contentType] = 'text/plain';
+    options.body = data;
+  }
+
+  console.log('callWebhook', event, url, data);
```
heichwald: curious why we sometimes use console and sometimes log.info

lib/services/notifications.js
```diff
@@ -0,0 +1,54 @@
...
+
+  restler.request(url, options).on('complete', function (result, response) {
+    log.info('called webhook', _.pick(response, ['status', 'statusText']), result);
```
heichwald: was thinking of maybe using https://github.com/request/request-promise to use promise everywhere ?

lib/services/notifications.js
TakenPilot: Ah, that should be removed.

lib/services/notifications.js
TakenPilot: Restler can be stubbed more easily.

Also, node-fetch is more generic and returns promises already.

lib/services/notifications.js
```diff
@@ -0,0 +1,54 @@
...
+ * @param {object} site
+ * @param {string} eventName
+ * @param {*} [data]
+ * @returns {function}
+ */
+function notify(site, eventName, data) {
+  const events = _.get(site, 'notify.webhooks');
+
+  if (events && _.isArray(events[eventName])) {
+    console.log('notify', eventName, events[eventName]);
```
heichwald: console here too

Issue #193, opened 2015-12-04
Don't PUT to @published@published?


yoshokatana:
When publishing :sparkles: again :sparkles: it's PUTting to `@published@published`

PR #194, opened 2015-12-04
Safe uri

TakenPilot:
Patch
- Do not add `@published` to `/uris/<page reference>` that is created when a page is published

Fixes https://github.com/nymag/amphora/issues/193


yoshokatana:
:+1:

Issue #198, opened 2015-12-09
POST /schedule doesn't create page@scheduled anymore


yoshokatana:
The `POST` request is returning 201 and has the correct data, and the entry in `/schedule` is correct, but `page@scheduled` doesn't exist

TakenPilot:
This was fixed by #199

PR #199, opened 2015-12-09
external-puts

TakenPilot:
Patch
- the `@scheduled` was not going to the right place after we made it external.


yoshokatana:
:+1:

yoshokatana:
fixes #198

PR #200, opened 2015-12-11
component server.js can return promises or objects

yoshokatana:
This gives an example of both


TakenPilot:
Returning a plain object isn't supported.  That it works in whatever example you're trying is probably just luck, and the code might change in the future that will prevent such things from working.

The docs are what we consider v1, so following those specs will ensure that your code keeps working until we hit v2.

PR #201, opened 2015-12-14
report memory leaks

TakenPilot:
Patch
- if any `_.memoize` function goes above 1000 internal items, log a warning a maximum of once a minute about the memory leak.

@chrismika @yoshokatana @gloddy @cruzanmo


yoshokatana:
:+1:

PR #207, opened 2016-01-12
Blank publish

TakenPilot:
Patch
- remove unused data from API tests
- add missing tests for versioned component data

PR #208, opened 2016-01-12
Blank publish allow

TakenPilot:
Dependent on this PR: https://github.com/nymag/amphora/pull/207

New Feature:
- When PUTing to `@published`, if no data is provided, will use `@latest` for pages and components

### DIFF
--- a/lib/services/components.js
+++ b/lib/services/components.js
@@ -269,10 +269,10 @@ function cascadingPut(uri, data, locals) {
 
 /**
  *
- * @param uri
- * @param data
- * @param locals
- * @returns {*}
+ * @param {string} uri
+ * @param {object} data
+ * @param {object} [locals]
+ * @returns {Promise}
  */
 function publish(uri, data, locals) {
   if (data && _.size(data) > 0) {
--- a/lib/services/components.js
+++ b/lib/services/components.js
@@ -268,6 +268,23 @@ function cascadingPut(uri, data, locals) {
 }
 
 /**
+ *
+ * @param uri
+ * @param data
+ * @param locals
+ * @returns {*}
+ */
+function publish(uri, data, locals) {
+  if (data && _.size(data) > 0) {
+    return cascadingPut(uri, data, locals);
+  } else {
+    return get(references.replaceVersion(uri), locals).then(function (data) {
+      return cascadingPut(uri, data, locals);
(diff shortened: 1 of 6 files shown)


test/mocha.opts
```diff
@@ -1,4 +1,4 @@
---reporter dot
+--reporter spec
```
cruzanmo: Interesting -- prettier?

cruzanmo:
Looks good. What is a use case for this feature?

lib/services/components.js
```diff
@@ -268,6 +268,23 @@ function cascadingPut(uri, data, locals) {
 }
 
 /**
+ *
+ * @param uri
+ * @param data
+ * @param locals
+ * @returns {*}
+ */
```
cruzanmo: Could add types to this JSDoc

test/mocha.opts
TakenPilot: Nah, I'll change this back.

lib/services/components.js
TakenPilot: Will add.

TakenPilot:
Ask Nelson about the use-case he needs.  But mostly just a shortcut, since passing empty data to a component is _always_ wrong, we're going to pretend that it has a use by assuming they meant `@latest`.  It'll allow us to write easier scripts to publish content without having to fetch `@latest` first.

test/mocha.opts
yoshokatana: b-but I like spec!

yoshokatana:
Yup, it'll allow us to do re-publishing of content a lot easier, and allow us to easily publish components. The most pressing use-case for this is of course publishing layouts, but it'll probably be used for more things.

lib/services/components.js
```diff
@@ -268,6 +268,23 @@ function cascadingPut(uri, data, locals) {
...
+function publish(uri, data, locals) {
+  if (data && _.size(data) > 0) {
+    return cascadingPut(uri, data, locals);
```
yoshokatana: niiiiiice

yoshokatana:
:+1:

yoshokatana:
even better! :+1: :+1: :shipit:

cruzanmo:
+1

PR #211, opened 2016-01-13
cascading-published-components

TakenPilot:
Patch
- Adds cascading ability to the shortcut of publishing components.

### DIFF
--- a/lib/services/components.js
+++ b/lib/services/components.js
@@ -8,6 +8,7 @@
 
 var _ = require('lodash'),
   control = require('../control'),
+  composer = require('../composer'),
   db = require('./db'),
   uid = require('../uid'),
   files = require('../files'),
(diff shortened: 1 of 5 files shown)


TakenPilot:
After some thought, this is actually a really bad idea:

```
dane [12:25 PM] 
@nelson   Causing components to cascade like this is kind of weird, considering people might start relying on it to publish a component ​_normally_​.
As in, someone edits the header, and then all the components in the header are auto published as well.

nelson [12:26 PM] 
Hmmmmm
Point
Hmm
Well, in most cases we'd want that

dane [12:27 PM] 
Even for components shared across layouts?

nelson [12:27 PM] 
Maybe if there's no data it should cascade, otherwise it should just publish the component (and data) you give it?

dane [12:28 PM] 
That's what I've built, but now we've given the lazy developer an easier way to publish a component without doing a GET.

nelson [12:28 PM] 
Haha
Hmm

dane [12:28 PM] 
So now components that are shared are being modified (and published) constantly.(edited)
This is starting to seem like a ​_really_​ bad idea.
I think I'd rather provide a way to fetch composed components.
And then they can publish ​_that_​.
After they've removed anything they don't want.
Or... hmm.
That could be a problem, because that'll include all the base components too.
No, this is ​_really really_​ bad, I think.
```

PR #212, opened 2016-01-20
better media adding when dealing with inline scripts / styles

yoshokatana:
This adjusts the `media` module's `findTop` and `findBottom` functions to better handle cases where components have inline `<script>` and `<style>` tags. The logic remains the same for all other situations.

* `findTop` (used to add `<link rel="stylesheet">`) will return the index right before the first inline style tag, if it exists
* `findBottom` (used to add `<script src>`) will return the index right before the first inline script tag, if it exists

### DIFF
--- a/lib/media.js
+++ b/lib/media.js
@@ -27,11 +27,17 @@ function wrapComments(list, name) {
  * @returns {number}
  */
 function findBottom(str) {
-  var index = str.lastIndexOf('</body>');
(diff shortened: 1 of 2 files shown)


TakenPilot:
Can you give examples of the use-case?

TakenPilot:
Also, why are we adding _after_ the inline reference?  Why not before?

PR #213, opened 2016-01-25
use X-Forwarded-Host for locals.url if exists

TakenPilot:
Patch
- `locals.url` was not respecting X-Forwarded-Host.  Now it does.


yoshokatana:
:+1:

PR #218, opened 2016-02-01
adds parsing of environment variables in yaml files

ughitsaaron:
adds a function to the files service for parsing yaml files using [template2env](https://github.com/ughitsaaron/template2env). this allows us to include environment variables in clay config files--making configuring clay for different environments (e.g., dev, staging, qa, and prod) much simpler.

### DIFF
--- a/lib/files.js
+++ b/lib/files.js
@@ -14,6 +14,7 @@ var _ = require('lodash'),
   path = require('path'),
   yaml = require('js-yaml'),
   pkg = require(path.resolve('package.json')),
+  temp2env = require('template2env'),
   req = require;
 
 /**
@@ -209,6 +210,15 @@ function getYaml(filename) {
 }
 
 /**
+ * @param {string} filename
+ * @return {string}
+ */
+function getYamlWithEnv(filename) {
+  let yamlParsed = temp2env(readFile(filename + '.yaml') || readFile(filename + '.yml'));
+  return yaml.safeLoad(yamlParsed);
+}
+
+/**
  * @param {object} value
  */
 function setPackageConfiguration(value) {
@@ -253,6 +263,7 @@ function getComponentPackage(name) {
 }
 
 exports.getYaml = control.memoize(getYaml);
+exports.getYamlWithEnv = control.memoize(getYamlWithEnv);
 exports.getFiles = control.memoize(getFiles);
 exports.getFolders = control.memoize(getFolders);
 exports.fileExists = control.memoize(fileExists);
--- a/lib/files.test.js
+++ b/lib/files.test.js
@@ -8,7 +8,8 @@ var _ = require('lodash'),
   yaml = require('js-yaml'),
   glob = require('glob'),
   lib = require('./' + filename),
-  pkg = require('../test/fixtures/config/package.json');
+  pkg = require('../test/fixtures/config/package.json'),
+  temp2env = require('template2env');
 
 describe(_.startCase(filename), function () {
   var req, sandbox;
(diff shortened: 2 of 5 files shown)


TakenPilot:
Just waiting for the coverage report.

lib/files.js
```diff
@@ -253,6 +263,7 @@ function getComponentPackage(name) {
 }
 
 exports.getYaml = control.memoize(getYaml);
+exports.getYamlWithEnv = control.memoize(getYamlWithEnv);
```
TakenPilot: This function seems like you just copy and pasted the function above it.  Did you just copy and paste the function above it?

Why do you want to memoize this function?  Do you see it being called repeatedly in short periods of time?

I think more thought might be needed here.

lib/files.test.js
```diff
@@ -164,12 +165,51 @@ describe(_.startCase(filename), function () {
 
     it('returns result', function () {
       var filename = 'some-name',
+        result = 'some result',
+        expected = 'some result';
+
+      yaml.safeLoad.returns(temp2env(result));
```
TakenPilot: I don't think this "temp2env" library is under test, so it shouldn't be run in our tests.

lib/files.test.js
```diff
@@ -164,12 +165,51 @@ describe(_.startCase(filename), function () {
...
+      var filename = 'some-name',
+        result = 'some ${FOO}',
+        expected = 'some result';
+
+      process.env.FOO = 'result';
+
+      yaml.safeLoad.returns(temp2env(result));
+
+      expect(fn(filename)).to.equal(expected);
+    });
```
TakenPilot: Again, it seems like you just copy and pasted nearby code.  That's very scary.

If you're not adding new logic that needs to be tested, then you're probably not doing something original.  More thought needs to be put into this.

lib/services/sites.js
```diff
@@ -53,8 +53,8 @@ function getSites() {
 
   files.getFolders(instanceSitesFolder).forEach(function (site) {
     var dir = path.join(instanceSitesFolder, site),
-      siteConfig = files.getYaml(path.join(dir, 'config')),
-      localConfig = files.getYaml(path.join(dir, 'local'));
+      siteConfig = files.getYamlWithEnv(path.join(dir, 'config')),
+      localConfig = files.getYamlWithEnv(path.join(dir, 'local'));
```
TakenPilot: Why does the sites service care that process.env is involved?  There are not env's being passed in.  There are no env's being passed out.

Encapsulation, and Single Responsibility Principle.

lib/services/sites.test.js
```diff
@@ -113,8 +113,8 @@ describe(_.startCase(filename), function () {
     it('throw error on missing host', function () {
       files.getFolders.returns(['a', 'b']);
       path.resolve.returns('z');
-      files.getYaml.onFirstCall().returns(getMockBadSite());
-      files.getYaml.onSecondCall().returns(getMockBadSite());
+      files.getYamlWithEnv.onFirstCall().returns(getMockBadSite());
+      files.getYamlWithEnv.onSecondCall().returns(getMockBadSite());
```
TakenPilot: This _looks_ like you just renamed a method, but the inputs and outputs are exactly the same.  Why is that?

TakenPilot:
I think this PR needs to be approached in a fundamentally different way.

I think that any PR that involved this much copy and paste should probably look elsewhere for where it's injecting its new logic.  Closing.

lib/files.js
ughitsaaron: > This function seems like you just copy and pasted the function above it. Did you just copy and paste the function above it?

No. One `getYamlWithEnv` uses `temp2env`. It seems to make sense to only use that in cases where you know you want environment variables to be included (e.g., config.yaml but not schema.yaml).

Memoizing the function makes sense for the way that the function is used in sites.js, no?

```
files.getFolders(instanceSitesFolder).forEach(function (site) {
  var dir = path.join(instanceSitesFolder, site),
    siteConfig = files.getYamlWithEnv(path.join(dir, 'config')),
    localConfig = files.getYamlWithEnv(path.join(dir, 'local'));
  // ...
});
```

PR #220, opened 2016-02-02
#getYaml checks filename & runs temp2env on "config" & "local"

ughitsaaron:
We want devs to be able to add environment variables to configuration files in order to simplify having different configurations in different environments, e.g., (local, QA, prod).

`getYaml` will check if the file it's reading is either "config" or "local" and will use [temp2env](https://github.com/ughitsaaron/template2env) to interpolate variable names in the config file (denoted with `${}`s) with environment variables.


TakenPilot:
+1 from Dane and Nelson

PR #224, opened 2016-02-10
add-batch-validation

TakenPilot:
Patch
- add batch validation, such that doubly-stringified objects will return an error.


yoshokatana:
+1

PR #229, opened 2016-02-19
better/sooner error messaging when templating eats an exception

TakenPilot:
Patch:

In certain cases, the templating engine has an error, and then it eats it, returning undefined.

Amphora's mediaMap should recognize that it is not being given what it expected, and say so.


yoshokatana:
:+1:

PR #233, opened 2016-02-24
log timeout and slow

TakenPilot:
Patch:
- Log when a particular server.js is slow.
- Fail operation when a particular server.js is twice as slow as that.

This is solving the case where we're not getting an idea of which component is causing a timeout (HTTP 504) on a proxy, so we're going to timeout before them.

Sites can set the timeouts to their own values to increase or decrease what is acceptable.

`require('amphora').components.setTimeoutConstant(10000)` means 
- GET: a component is warned of being slow at 10000ms, and timeouts at 20000ms.
- PUT: a component is warned of being slow at 25000ms, and timeouts at 50000ms.

### DIFF
--- a/lib/services/components.js
+++ b/lib/services/components.js
@@ -6,7 +6,7 @@
 
 'use strict';
 
-let timeoutConstant = 2000;
+let timeoutConstant = 4000;
 const _ = require('lodash'),
   control = require('../control'),
   db = require('./db'),
--- a/lib/services/components.js
+++ b/lib/services/components.js
@@ -6,11 +6,14 @@
 
 'use strict';
 
-var _ = require('lodash'),
+let timeoutConstant = 2000;
+const _ = require('lodash'),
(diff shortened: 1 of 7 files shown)


lib/services/components.js
```diff
@@ -18,7 +21,23 @@ var _ = require('lodash'),
   path = require('path'),
   glob = require('glob'),
   referenceProperty = '_ref',
-  mediaDir = path.join(process.cwd(), 'public');
+  mediaDir = path.join(process.cwd(), 'public'),
+  timeoutGetCoefficient = 2,
+  timeoutPutCoefficient = 5;
```
TakenPilot: timeoutConstant * timeoutCoefficient => the interval we're willing to wait

TakenPilot:
This code works locally through an entire import of scienceofus.

yoshokatana:
:+1:

PR #236, opened 2016-03-01
moves addQueryParams from addControllerRoutes to addSites

ughitsaaron:
params need to be added in `addSite` before `addControllerRoutes` in order to expose query params without breaking edits when `req.query.edit` is `true`.


yoshokatana:
Does this need a test / changes to the current tests?

ughitsaaron:
Not sure. Tests pass and 100% is still covered. Not sure what/how it would be tested?

PR #239, opened 2016-03-03
expose decomposing-components; remove lodash-ny-util

TakenPilot:
Patch:
- expose `composer` service
- remove `lodash-ny-util` dependency


yoshokatana:
:+1:

PR #240, opened 2016-03-03
esnext

TakenPilot:
Patch:
- add more restrictions to prefer the new ESNext syntaxes where possible.


cruzanmo:
+1. In circle we trust.

PR #242, opened 2016-03-07
Added assetPath logic for 'root' site scenarios

larrychevres:
This fixes style and js references for sites like "vulture" and "grubstreet" where site path is "/" and assetPath setting is "/vulture"

### DIFF
--- a/lib/services/components.js
+++ b/lib/services/components.js
@@ -451,12 +451,12 @@ function getIndices(ref, data) {
 function getStyles(name, slug) {
   const site = siteService.sites()[slug],
     assetDir = site && site.assetDir || mediaDir,
-    assetDirLen = assetDir.length,
-    assetPath = site && site.assetPath || '';
+    assetDirLen = assetDir.length;
+  let assetPath = site && site.assetPath || '';
 
-   if (site.assetPath != site.path) {
-      assetPath = site && site.path || '';
-    }
+  if (site && site.path && site.assetPath !== site.path) {
+    assetPath = site.path;
+  }
 
   return _.map(_.filter([
     path.join(assetDir, 'css', name + '.css'),
@@ -473,12 +473,12 @@ function getStyles(name, slug) {
 function getScripts(name, slug) {
   const site = siteService.sites()[slug],
     assetDir = site && site.assetDir || mediaDir,
-    assetDirLen = assetDir.length,
-    assetPath = site && site.assetPath || '';
+    assetDirLen = assetDir.length;
+  let assetPath = site && site.assetPath || '';
 
-   if (site.assetPath != site.path) {
-      assetPath = site && site.path || '';
-    }
+  if (site && site.path && site.assetPath !== site.path) {
+    assetPath = site.path;
+  }
 
   return _.map(_.filter([
     path.join(assetDir, 'js', name + '.js'),
--- a/lib/services/components.js
+++ b/lib/services/components.js
@@ -454,6 +454,10 @@ function getStyles(name, slug) {
     assetDirLen = assetDir.length,
     assetPath = site && site.assetPath || '';
 
+   if (site.assetPath != site.path) {
+      assetPath = site && site.path || '';
+    }
+
   return _.map(_.filter([
     path.join(assetDir, 'css', name + '.css'),
     path.join(assetDir, 'css', name + '.' + slug + '.css')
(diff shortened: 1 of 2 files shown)


lib/services/components.js
```diff
@@ -469,8 +473,12 @@ function getStyles(name, slug) {
...
   const site = siteService.sites()[slug],
     assetDir = site && site.assetDir || mediaDir,
-    assetDirLen = assetDir.length,
-    assetPath = site && site.assetPath || '';
+    assetDirLen = assetDir.length;
+  let assetPath = site && site.assetPath || '';
+
+  if (site && site.path && site.assetPath !== site.path) {
+    assetPath = site.path;
+  }
```
TakenPilot: Looking at this again, I can see why it is a bit weird. It's saying:  "If there is a site that has a path, and if the path is different than the assetPath, use the path for the assetPath."
```
 if (site && site.path && site.assetPath !== site.path) {
    assetPath = site.path;
 }
```

is the same as 

```
 if (site && site.path) {
    assetPath = site.path;
 }
```

because you're saying assetPath should be the same as path if path exists.  You could combine that with the statement above to create

```
let assetPath = site && (site.path || site.assetPath || '');
```

Is that what we really want?

It seems more likely that this is a configuration error, doesn't it?

lib/services/components.test.js
```diff
@@ -666,12 +682,28 @@ describe(_.startCase(filename), function () {
...
+      expect(fn('name', 'slug')).to.deep.equal(['/scienceofus/css/name.css', '/scienceofus/css/name.slug.css']);
+    });  
+
+    it('accepts good component with slug (with slug file) with assetDir and assetPath and path', function () {
+      siteService.sites.returns({slug: {assetDir: 'someAssetDir', path: '/', assetPath: '/vulture'}});
+      files.fileExists.withArgs('someAssetDir/css/name.css').returns(true);
+      files.fileExists.withArgs('someAssetDir/css/name.slug.css').returns(true);
+
+      expect(fn('name', 'slug')).to.deep.equal(['/css/name.css', '/css/name.slug.css']);
+    });
```
TakenPilot: This is a good test because it shows what the expectations are.

```
assetDir: 'someAssetDir'
path: '/'
assetPath: '/vulture'
```

With this configuration, I think people would expect the assets to be found at `/vulture`, not `/`.  However, the test is expecting the assets at `/`.

Could we just set the assetPath to `/` instead?

lib/services/components.js
larrychevres: Yeah, agree it is all a bit awkward. I just kept tweaking it to meet the tests which I could not get to pass with config changes alone.  These two tests are what are needed to run all current sites.

I think there are 2 real-world cases (and 1 not so real, but possible via current config options):
1) 
site.path = '/scienceofus'
site.assetPath = '/scienceofus'
sample url: /scienceofus/css/public-login.css'
filepath: '/css/simple-header.scienceofus.css'
2) 
site.path = '/'
site.assetPath = '/vulture'
sample url: /css/simple-header.css'
filepath: '/css/simple-header.vulture.css'
3) 
site.path = '/scienceofus'
site.assetPath = '/scienceofusAssetPath'
sample url: /scienceofus/css/public-login.css'
filepath: '/css/simple-header.scienceofus.css' ????

TakenPilot: I think I understand.

For number two, why is it good to have an asset path like that?  Wouldn't that cause `http://vulture.com/vulture/css/simple-header.css` ?

There is this code in sites.js:
```
    siteConfig.slug = site;
    siteConfig.path = normalizePath(siteConfig.path);
    siteConfig.dir = dir;
    siteConfig.assetPath = normalizePath(siteConfig.assetPath);
    siteConfig.assetDir = normalizeDirectory(siteConfig.assetDir || 'public');
    siteConfig.prefix = siteConfig.path && siteConfig.path.length > 1 ? siteConfig.host + siteConfig.path : siteConfig.host;
```

which I think means the configs were meant to be something like this for vulture:
```
proto: (defaults to http)
port: (defaults to env.PORT)
slug: (defaults to name of the folder)
host: vulture.com
path: (defaults to /)
assetPath: (defaults to /)
assetDir: (defaults to public)
```

At least, that was the intention.

larrychevres: hmm, so given above, I don't think I was able to get vulture to properly serve its assets using every combination I could on those variables (though I did not mess with assetDir), hence my change.  I feel like somewhere the slug or assetPath are being used incorrectly, requiring my update but could be wrong

PR #243, opened 2016-03-08
PUT @published with empty data

yoshokatana:
Because of the recent changes to publishing, we can remove some extraneous functionality from the scheduler.

Now it will PUT to `whatever@published` with empty data, rather than fetching the latest data first. This removes some complexity from the scheduler, and has the added bonus of allowing us to schedule the publishing of arbitrary components (which will do a recursive publish of their child components)

### DIFF
--- a/lib/services/schedule.js
+++ b/lib/services/schedule.js
@@ -31,11 +31,10 @@ function publishExternally(url) {
     const latest = references.replaceVersion(url),
       published = references.replaceVersion(url, 'published');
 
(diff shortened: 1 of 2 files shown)


TakenPilot:
+1

yoshokatana:
:-1: publishing pages doesn't just use the latest data, so this breaks page scheduling

yoshokatana:
Investigating solutions now...

yoshokatana:
Ok, the solution is https://github.com/nymag/sites/pull/705

PR #244, opened 2016-03-09
include port in locals.url

TakenPilot:
Patch
- Currently, port is not included in locals.url.  Now it will be.


ughitsaaron:
+1

PR #246, opened 2016-03-10
deny-trailing-slash

TakenPilot:
Patch
- Explicitly 400 when client tries to PUT or GET from /components/, /pages/, or /uris/ with an id with a trailing slash.

`a/components/b/instances/c/` => `400 Trailing slash on RESTful id in URL is not acceptable`
`a/components/b/instances/c` => `200 Success`


yoshokatana:
:+1:

PR #247, opened 2016-03-21
allow deletes of versions

TakenPilot:
Allow someone to delete versioned components, like `article/components/abc@published`

This is needed when we want to go back to the state of "this thing has not been published before".


yoshokatana:
oh that's useful :+1:

byronhulcher:
+1

PR #252, opened 2016-04-06
allow PUTs to /components/name/instances/id.html

yoshokatana:
This allows us to send PUT requests to instances with extensions, e.g. `.json` or `.html`. As with the equivalent GET requests, the extension determines the type of the returned data.

This is useful for Kiln (or other editors) to update components and re-render them without having to do a separate call to grab the rendered HTML after saving.


yoshokatana:
TODO: In the future we should probably also allow POSTs to return html

byronhulcher:
:+1:

PR #254, opened 2016-04-08
don't set read-only data in html-composer

yoshokatana:


yoshokatana:
I'm merging this so we can test on QA and make sure it works. If it doesn't, we'll revert this PR.

PR #256, opened 2016-04-11
deprecate jshint + jscs, use our standard eslint config

yoshokatana:
This PR brings amphora in line with our standard eslint configuration, allowing us to stop dealing with jscs and jshint


yoshokatana:
Freezing the object again brings the codebase back to the point where everything started breaking, so I can start diagnosing the issues and try to fix them.

byronhulcher:
+1

PR #262, opened 2016-04-19
Media stream (Part 1)

jonwinton:
This PR inlines the styles and scripts into the HTML response. It concatenates all the CSS files and script files into big strings and injects them into the proper part of the page.

Trello card: https://trello.com/c/jpTJ7YEQ/279-script-style-streaming

Next up: actually streaming the response. Still looking into the best way to do this.


lib/files.test.js
```diff
@@ -335,4 +335,44 @@ describe(_.startCase(filename), function () {
       expect(fn(name)).to.equal(result);
     });
   });
+
+
+
+
+
+
+
```
yoshokatana: eeeeempty spaaaaaaace

lib/html-composer.js
```diff
@@ -115,7 +115,8 @@ function renderTemplate() {
     }
 
     return Promise.resolve(multiplex.render(template, data))
-      .then(media.append(data));
+      .then(media.append(data))
+      .then(html => html); // media.append() returns Promise
```
yoshokatana: does `media.append()` not return the html?

lib/media.js
```diff
@@ -14,24 +16,16 @@ function flatMap(list, fn) {
...
-  list.push('<!-- ' + name + ' End -->\n');
-}
-
-/**
  * @param {string} str
  * @returns {number}
  */
 function findBottom(str) {
-  let index = str.lastIndexOf('</body>');
+  var index = str.lastIndexOf('</body>');
```
yoshokatana: why `var` instead of `let`?

lib/media.js
```diff
@@ -40,11 +34,12 @@ function findBottom(str) {
  * @returns {number}
  */
 function findTop(str) {
-  let index = str.indexOf('</head>');
+  var index = str.indexOf('</head>');
```
yoshokatana: why `var` instead of `let`?

lib/media.js
```diff
@@ -57,42 +52,100 @@ function findTop(str) {
...
+  var allPromises = [],
+    rootDir = process.cwd();
+
+  fileArray.forEach(file => {
+    allPromises.push(files.readFilePromise(path.join(rootDir, targetDir, getFileName(file, filePrefix))));
+  });
+
+  return Promise.all(allPromises)
+    .catch(err => {
+      throw err;
```
yoshokatana: this should throw higher up in the stack already, shouldn't it?

lib/media.js
```diff
@@ -57,42 +52,100 @@ function findTop(str) {
...
+ */
+function combineStyles(stylesArr) {
+  if (!stylesArr || !stylesArr.length > 0) {
+    return false;
+  }
+
+  // If there are styles, retrieve contents
+  return getContentsOfFiles(stylesArr, 'public/css', '/css/')
+    .then(data => {
+      return '<style>' + data.join('') + '</style>';
```
yoshokatana: this function could be on one line. generally we like to only use fat arrows for single-line functions, and the full `function () {}` syntax for multiline

lib/media.js
```diff
@@ -57,42 +52,100 @@ function findTop(str) {
...
+ */
+function combineScripts(scriptsArr) {
+  if (!scriptsArr || !scriptsArr.length) {
+    return false;
+  }
+
+  // If there are scripts, retrieve contents
+  return getContentsOfFiles(scriptsArr, 'public/js', '/js/')
+    .then(data => {
+      return '<script type="text/javascript">' + data.join('') + '</script>';
```
yoshokatana: same with this

lib/media.js
```diff
@@ -57,42 +52,100 @@ function findTop(str) {
...
+    .then(data => {
+      return '<style>' + data.join('') + '</style>';
+    });
+}
+
+/**
+ * Concatenate all the scripts together in one `<script>` tag
+ *
+ * @param  {array} scriptsArr The array of all the script files needed for a page
+ * @return {Promise}
```
yoshokatana: this looks like it returns `Promise` or `boolean`, no?

lib/media.js
```diff
@@ -113,15 +168,16 @@ function append(data) {
...
-
-    if (mediaMap.scripts && mediaMap.scripts.length > 0) {
-      html = appendMediaToBottom(mediaMap, html);
-    }
+    // Get the combined styles and scripts
+    stylesPromise = combineStyles(mediaMap.styles),
+    scriptsPromise = combineScripts(mediaMap.scripts);
 
-    return html;
+    return Promise.all([stylesPromise, scriptsPromise])
```
yoshokatana: since you're grabbing specific properties, you could use `bluebird.props`

lib/html-composer.js
jonwinton: Nope, `media.append()` returns a `Promise`.

lib/media.js
jonwinton: I accidentally deleted this line in one commit and then re-added it. I forgot that we can render one component if we use it's unique URL. 

But, also, there's no real need for `let` since `index` is scoped globally to this function. Using `let` doesn't actually serve a purpose in this case.

lib/media.js
jonwinton: See above comment. Again, just accidentally deleted in a previous commit.

lib/media.js
jonwinton: Hmm the only step that's higher in the stack is the actual reading of the file's contents, and I didn't want to throw it there because each module that uses the function should be able to determine how the error is handled.

lib/media.js
jonwinton: 👍

lib/media.js
jonwinton: 👍

lib/media.js
jonwinton: 👍

lib/media.js
jonwinton: Because I went the route of ES6 Promises I was thinking it would be better to stick with that implementation.

lib/html-composer.js
yoshokatana: yeah, but it returns a promise that returns html, right?

lib/media.js
yoshokatana: eh, ok

lib/html-composer.js
jonwinton: Ohhh dang it, this was a change from before I updated the scripts with a `Promise`. You're right, good catch. Can remove that line.

lib/media.js
yoshokatana: Ah makes sense.

lib/media.js
yoshokatana: We already have bluebird as a dep tho 😄

lib/files.test.js
```diff
@@ -335,4 +335,31 @@ describe(_.startCase(filename), function () {
...
-});
+
+  describe('readFilePromise', function () {
+    const fn = lib[this.title];
+
+    it('returns file contents', function () {
+      const result = 'result',
+        name = 'article.css';
+
+      sandbox.stub(fs, 'readFile', function (path, x, callback) {
```
dmabuada: what is x exactly? kind of confusing.

lib/media.js
```diff
@@ -57,50 +52,106 @@ function findTop(str) {
...
+ * Retrieve the contents of a file
+ *
+ * @param  {array} fileArray   An array of file names
+ * @param  {string} targetDir  The directory to retrieve files from
+ * @param  {string} filePrefix The string that comes right before the file name
+ * @return {Promise}
+ */
+function getContentsOfFiles(fileArray, targetDir, filePrefix) {
+  var allPromises = [],
+    rootDir = process.cwd();
```
dmabuada: this could be more aptly named "currentDir", because semantics are fun. unless you're assuming that current dir is always root dir?

lib/media.js
```diff
@@ -57,50 +52,106 @@ function findTop(str) {
...
+}
+
+/**
+ * Concatenate all the scripts together in one `<script>` tag
+ *
+ * @param  {array} scriptsArr The array of all the script files needed for a page
+ * @return {Promise | boolean}
+ */
+function combineScripts(scriptsArr) {
+  if (!scriptsArr || !scriptsArr.length) {
```
dmabuada: length > 0 ?

lib/files.js
```diff
@@ -275,6 +275,23 @@ function getComponentPackage(name) {
   return componentPath && tryRequireEach([componentPath + '/package.json']);
 }
 
+/**
+ * Returns a promise representing the retrieval of content from a file
```
davidandrewfrey: Good description, but just needs to follow JSDoc formatting:
http://usejsdoc.org/

lib/media.js
dmabuada: also these 2 functions essentially do the same thing. could there be a function that handles both?

lib/media.js
dmabuada: ^^ 👍

lib/files.test.js
davidandrewfrey: Agreed. Especially hard if the code is self-documenting.

jonwinton: Yeah, this is weird, sorry about that. The actual [`fs.readFile`](https://nodejs.org/api/fs.html#fs_fs_readfile_file_options_callback) function accepts an optional options object **or** string as a second argument (say that five times fast). In the actual function I pass `utf-8` as the encoding option, so for the test Morgan and I figured out we just needed a placeholder. I could name that differently though. I think we were just trying to debug the tests quickly and I forgot to re-name.

lib/media.js
jonwinton: I called it `rootDir` because it's running from the root of the `sites` repo which has access to all the files. Should it be something like `projectRoot`? Or `projectDir`? I defer to your naming because I'm the worst at naming.

lib/media.js
jonwinton: - Need to remove the `> 0`. Doesn't break, but unnecessary

- I considered merging them into one, but I also kind of like the idea of keeping style and script concatenation along two separate paths incase we ever want to perform secondary operations on one of them. I was thinking about how we're going to end up injecting scripts for one-off interactives and it might be necessary to further modify the scripts...? Maybe I'm thinking too much into it and that'll be another task. Thoughts?

lib/files.test.js
dmabuada: a rename would be nice, to prevent further head tilts.

lib/media.js
davidandrewfrey: Agree with keeping these two separated as they're affecting separate resource types.

lib/media.js
dmabuada: currentDir probably still makes sense that's what process.cwd is (literally)!

https://nodejs.org/api/process.html#process_process_cwd

lib/media.js
davidandrewfrey: Regarding the array check above, feel like we need to come together as a group for a consensus for this instant and other things we do. Especially if one of our goals is to have our code be uniform enough to look like it was all written by one person.

dmabuada: hmmm, what kind of secondary operations? it's possible. it just doesn't seem very DRY as is, but maybe a second opinion is a good idea.

dmabuada: also why does the first function have length > 0?

jonwinton: I think that was a copy paste error...... 👎  I'm going to update that right now. 


But as for secondary operations, what if we wanted to fetch the JS content of an interactive from some external source? Aaron and I were just spitballing about how it would be cool to have a repo for each interactive, they push compiled code to a `release` branch or something, there's a module in Clay which is just given the URL of a repo, and, _finally_, in compilation we can grab those files.

Granted, that idea has some holes in it and it the result of 3 minutes of conversation, but if there's some abnormal method of injecting styles/scripts for one-off interactives then we might want them separate? This stems from the issue of making each interactive an individual component and then cluttering up our components directory. 

Again, this is a topic for a different conversation, but that sort of flexibility might be nice in the future? Maybe I just took us on a bad tangent...

dmabuada: gonna defer to @yoshokatana or @gloddy on this one

lib/media.test.js
```diff
@@ -58,52 +62,98 @@ describe(_.startCase(filename), function () {
...
 
-    it('adds to bottom of root', function () {
-      components.getScripts.onCall(0).returns(['/js/a.js']);
+    describe('with scripts', function () {
+      function resolveBody(html) {
+        expect(html).to.deep.equal(componentScriptHtml);
+      }
+
+      function resolveSection(html) {
+        expect(html).to.deep.equal(componentScriptSection);
```
davidandrewfrey: Broken up nicely!

davidandrewfrey:
👍

dmabuada:
much cleaner. +1

gloddy:
👍

PR #263, opened 2016-04-21
Allow a blank path and assetPath

byronhulcher:
Will support sites that exist only with a a domain.

### DIFF
--- a/lib/services/sites.js
+++ b/lib/services/sites.js
@@ -11,11 +11,13 @@ const _ = require('lodash'),
  * @returns {string}
  */
 function normalizePath(urlPath) {
-  if (!urlPath || urlPath === '/') {
+  if (!urlPath) {
     return '';
+  // make sure path starts with a /
   } else if (_.first(urlPath) !== '/') {
     urlPath = '/' + urlPath;
   }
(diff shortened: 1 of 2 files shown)


cruzanmo:
From slack conversation: seems that `site.assetPath` and `site.path` should follow the same convention.

cruzanmo:
This fix can be made to the normalizePath function itself https://github.com/nymag/amphora/blob/allow-blank-assetPath/lib/services/sites.js#L15

lib/services/sites.js
```diff
@@ -5,23 +5,18 @@ const _ = require('lodash'),
...
- * Normalize path to always exist, always start with a slash, and never end with a slash.
+ * Normalize path to never end with a slash.
  *
  * @param {string} urlPath
  * @returns {string}
  */
 function normalizePath(urlPath) {
-  if (!urlPath) {
-    urlPath = '/';
+  if (!urlPath || urlPath === '/') {
```
cruzanmo: `urlPath === '/'` case is handled by `trimRight` so do not need it here.

cruzanmo:
+1

PR #264, opened 2016-04-21
error -> verbose on bootstrap

ughitsaaron:
Change level of "No bootstrap found" message.


larrychevres:
I think it should be info level at very least, potentially warn.

ughitsaaron:
@larrychevres @cruzanmo and I were discussing it and i think we wanna leave this as is for now because it relates to a wider conversation clay & bootstraps. closing.

PR #265, opened 2016-04-29
Inline font file

jonwinton:
Extending upon this card: https://trello.com/c/jpTJ7YEQ/279-script-style-streaming

This PR is for inlining fonts files. Right now fonts are included as a separate request in the `layout` component, this will allow us to remove that `link`.

### DIFF
--- a/lib/html-composer.js
+++ b/lib/html-composer.js
@@ -104,9 +104,11 @@ function applyOptions(options, res) {
 
 /**
  * Render a template based on some generic data provided.
+ *
+ * @param {Object} res
  * @returns {Function}
  */
-function renderTemplate() {
+function renderTemplate(res) {
   return function (data) {
     const template = components.getTemplate(data.template, defaultHtmlTemplateName);
 
@@ -115,7 +117,7 @@ function renderTemplate() {
     }
 
     return Promise.resolve(multiplex.render(template, data))
-      .then(media.append(data));
+      .then(media.append(data, res.locals));
   };
 }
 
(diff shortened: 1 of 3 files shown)


lib/html-composer.js
```diff
@@ -104,9 +104,11 @@ function applyOptions(options, res) {
 
 /**
  * Render a template based on some generic data provided.
+ *
+ * @param {Object} res
  * @returns {Function}
  */
-function renderTemplate() {
+function renderTemplate(res) {
```
byronhulcher: kinda pendantic but I'd rather you pass in the `locals` property, since that's what you're really after.

lib/media.js
```diff
@@ -114,48 +114,69 @@ function wrapWithTags(string, tag) {
...
+ */
+function getFonts(site) {
+  var pathToFontCSS = path.join(currentDir, site.assetDir, 'fonts', `combined-${site.slug}.css`);
+
+  // Return a promise for reading the file's contents
+  return files.readFilePromise(pathToFontCSS)
+    .catch(function () {
+      // If there was an error reading from the file
+      // because it doesn't exist, return an empty string
+      return '';
```
byronhulcher: Would it be useful if it output a CSS comment perhaps? Maybe for debugging purposes?

byronhulcher:
Two super minor comments, but other than that its 👍  and could be merged

yoshokatana:
Per our conversation, Jon's gonna do this in sites instead (using the resolveMediaMap functionality to keep amphora generic)

jonwinton:
Womp womp gonna delete this 👎

PR #268, opened 2016-05-12
Add CLAY_ACCESS_KEY support for restful calls

larrychevres:
Add support to pass access key that clay instances may enforce for some HTTP operations.  Specifically fixes scheduling, where publishes are achieved through a self PUT.  Requires implementation in sites project.


yoshokatana:
👍

PR #269, opened 2016-05-16
simple key mapping for page areas

yoshokatana:
**BREAKING CHANGE**

This replaces the (complicated) insertion of page data into layouts with a simple 1:1 key mapping. The old way was extremely flexible, but complicated and slow (and it turns out was more than we actually needed). The new way allows us to bake some assumptions into our data and kiln, and lays the foundation for switching layouts.

_Here's what the **old** data looked like:_

```yaml
# page
pageRef:
  layout: /layout/ref
  areaA: /a/string/ref
  areaB:
    - /another/ref
    - /a/third/ref

# layout
layoutRef:
  propA: ['areaA']
  propB:
    -
      _ref: /some/layout/component
    - areaB
```
_Here's what the **new** data looks like:_

```yaml
# page
pageRef:
  layout: /layout/ref
  propA:
    - /a/string/ref # all page areas look the same
  propB:
    - /another/ref
    - /a/third/ref

# layout
layoutRef:
  propA: propA # layout prop and page prop have the same name, allowing for easy insertion and switching
  propB: propB
  propC:
    -
      _ref: /some/layout/component
```

### DIFF
--- a/lib/html-composer.js
+++ b/lib/html-composer.js
@@ -166,66 +166,20 @@ function refToObj(ref) {
 }
 
 /**
- * Clear and replace original list with items in-place (which may have different length than original)
- *
- * @param {Array} list  Array to clear and replace
- * @param {Array} items  Items to put into the list.
- */
-function replaceItems(list, items) {
-  list.length = 0;
-  list.push.apply(list, items);
-}
-
-/**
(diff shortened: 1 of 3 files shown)


byronhulcher:
👍

PR #274, opened 2016-06-01
301 redirect when uri matches another uri

yoshokatana:
When a request comes in for a uri and its data in `/uris/` matches _another_ uri, it will send a 301 redirect rather than recursively looking up uris to serve the page.

From a user's perspective, this will still load the page they were expecting, but the url in their address bar will be the updated url (which corresponds to the uri that points _directly_ to the page).

### DIFF
--- a/lib/html-composer.js
+++ b/lib/html-composer.js
@@ -239,8 +239,8 @@ function renderUri(uri, res) {
     if (route.isUri) {
       let newBase64Uri = _.last(result.split('/')),
         newUri = new Buffer(newBase64Uri, 'base64').toString('utf8'),
-        newPath = url.parse('http://' + newUri).path,
-        newUrl = `http://${res.req.get('host')}${newPath}`;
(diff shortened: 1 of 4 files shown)


lib/html-composer.js
```diff
@@ -235,7 +236,17 @@ function renderUri(uri, res) {
       throw new Error('Invalid URI: ' + uri + ': ' + result);
     }
 
-    return route.html(result, res);
+    if (route.isUri) {
+      let newBase64Uri = _.last(result.split('/')),
+        newUri = new Buffer(newBase64Uri, 'base64').toString('utf8'),
+        newPath = url.parse('http://' + newUri).path,
+        newUrl = `http://${res.req.get('host')}${newPath}`;
```
cruzanmo: Assumes `http://`

yoshokatana: updating

byronhulcher:
👍

PR #275, opened 2016-06-02
allow oldUrls array in page data

yoshokatana:
Published pages contain the "canonical" url, but they should also store a simple array of _all_ the urls that have been added to it. This allows 3rd party services to easily know when pages have been re-published, and allows us to easily add redirects when re-publishing.

They are listed oldest to newest, like so:

```json
{
  "url": "http://nymag.com/scienceofus/2016/06/new-slug.html",
  "oldUrls": [
    "http://nymag.com/scienceofus/2016/06/old-slug.html",
    "http://nymag.com/scienceofus/2016/06/new-slug.html"
  ]
}
```

### DIFF
--- a/lib/services/pages.js
+++ b/lib/services/pages.js
@@ -136,8 +136,8 @@ function replacePageReferenceVersions(data, version) {
 
     if (references.isUri(value)) {
       result = replaceVersion(value);
-    } else if (_.isArray(value) && key !== 'oldUrls') {
-      // oldUrls is an array of old page urls. they're not component refs and shouldn't be versioned
+    } else if (_.isArray(value) && key !== 'urlHistory') {
+      // urlHistory is an array of old page urls. they're not component refs and shouldn't be versioned
       result = _.map(value, replaceVersion);
     } else {
       result = value;
--- a/lib/services/pages.js
+++ b/lib/services/pages.js
@@ -131,12 +131,13 @@ function applyBatch(site) {
 function replacePageReferenceVersions(data, version) {
   const replaceVersion = _.partial(references.replaceVersion, _, version);
 
-  return _.mapValues(data, function (value) {
+  return _.mapValues(data, function (value, key) {
     let result;
 
     if (references.isUri(value)) {
       result = replaceVersion(value);
-    } else if (_.isArray(value)) {
+    } else if (_.isArray(value) && key !== 'oldUrls') {
+      // oldUrls is an array of old page urls. they're not component refs and shouldn't be versioned
       result = _.map(value, replaceVersion);
     } else {
       result = value;
--- a/lib/services/references.js
+++ b/lib/services/references.js
@@ -164,7 +164,7 @@ function getUriPrefix(uri) {
  * @returns {object}
  */
 function omitPageConfiguration(pageData) {
-  return _.omit(pageData, ['layout', 'url', 'oldUrls', 'lastModified', 'priority', 'changeFrequency']);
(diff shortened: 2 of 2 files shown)


saml:
Current url is included in oldUrls?  such as   http://nymag.com/scienceofus/2016/06/new-slug.html  
Maybe rename to allURLs?

yoshokatana:
Hmm, good point.

saml:
On a related note, it'd be useful to include `id` field in page representation, since it looks like `url` isn't stable. Something like GUID in RSS feeds.  

3rd party vendors that listen to amphora page notification event can use stable id of the page instead of history of changing URLs of the page. They could just update their datastore with given page id, instead of deleting multiple items listed in oldUrls and adding one item with url (assuming they have built their datastore where key is page's url).

cruzanmo:
@saml the page URI is stable. Perhaps we could use that?

saml:
@cruzanmo a page can have many (public) urls.

For example, `/asite/pages/a`  is stable. This key won't change in amphora.
But there are  `/asite/uris/a`, `/asite/uris/b`, `/asite/uris/c`, ..   that resolve to `/asite/pages/a`.
You can use any of those public uris to access `/asite/pages/a`.

It'd be nice to include the stable (private?) key,  `/asite/pages/a` in the page json as well.

yoshokatana:
We don't include the page uri in the page data itself because anyone who's doing a request for the page already knows the uri. That being said, the notifications service (which fires the webhooks) should somehow include the uri in the stuff it's sending, since it's not requesting page data from the uri.

cruzanmo:
To keep it RESTful, we could have the page URI in a header, i.e. a link header https://www.w3.org/wiki/LinkHeader

yoshokatana:
@cruzanmo @saml let's continue this discussion in this ticket https://github.com/nymag/amphora/issues/276

cruzanmo:
Adding URL history to the body adds convenience, but might also be also be adding complexity. A third-party service should only need a stable id to know if a page is being re-published.

yoshokatana:
@cruzanmo how would a 3rd party service know that the url has _changed_, though?

yoshokatana:
(also, without some kind of record, even our internal publishing service has no idea if a page url has changed)

saml:
@yoshokatana I'm not sure if 3rd party service is concerned with history or change of public URL as long as they have ID of the page. When a 3rd party service is using url as id, it's nice for amphora to provide history of url so that the 3rd party can know if current publish event is for one of the known pages, not a new page. And it can update the existing page instead of creating a new one.
But if 3rd party service uses id as id, it does not care if current publish event is new or republish of an existing page. It can diff the change that's included in this publish event against its datastore, detect url change, and do whatever it wants to do.

I think the problem is that amphora currently includes only url in publish event that could be used to identify the page being published. 
Including stable id in publish event would be great, in addition to providing url history.

yoshokatana:
> it's nice for amphora to provide history of url

This is exactly what this PR does...

yoshokatana:
The discussion of whether notifications should include a uri is separate from this pull request, and should be done here → https://github.com/nymag/amphora/issues/276

cruzanmo:
I do not think it is necessary to include the URL history, if we resolve issue #276.

yoshokatana:
This PR really shouldn't depend on resolving that issue, though. They're completely unrelated.

yoshokatana:
👍  from our irl discussion

PR #277, opened 2016-06-24
Allow a site to supply its own chain of responsbility for publishing a page

byronhulcher:
This is work in advance of allowing each site to custom define its own url structure.  This allows each site to supply its own hierarchy of rules to determine what url to generate when publishing a page. 

You can observe the changes to our implentation that will be required with this update here at: https://github.com/nymag/sites/pull/1236

### DIFF
--- a/lib/services/pages.js
+++ b/lib/services/pages.js
@@ -242,8 +242,10 @@ function publish(uri, data, locals) {
     prefix = references.getPagePrefix(uri),
     site = getSite(prefix, locals),
     timeoutLimit = timeoutConstant * timeoutPublishCoefficient;
-
-  return getPublishData(uri, data).then(function (pageData) {
+  return getPublishData(uri, data)
+    .then(function(pageData){
+      return _.isFunction(site.publishPage) ? site.publishPage(uri, pageData, locals) : Promise.resolve(pageData);
+    }).then(function (pageData) {
     const published = 'published',
       publicUrl = pageData.url,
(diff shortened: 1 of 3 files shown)


lib/services/pages.js
```diff
@@ -243,16 +243,24 @@ function publish(uri, data, locals) {
...
-    const published = 'published',
-      publicUrl = pageData.url,
-      componentList = references.getPageReferences(pageData);
-
-    if (!references.isUrl(publicUrl)) {
-      throw new Error('Client: Page must have valid url to publish.');
-    }
+  return getPublishData(uri, data)
+    .then(function (pageData) {
+      if ( _.isArray(site.publishingChain) ) {
```
yoshokatana: "loop through the array of publishing functions, finding the first one that resolves"

"this is a variant of the Chain of Responsibility pattern, which we're calling Reject Quickly/Resolve Slowly. functions should reject quickly if the data doesn't match the format they want, and _afterwards_ resolve the generated url if the data _does_ match"

etc

byronhulcher: Added a comment, lemme know watchu think

byronhulcher:
I gotta look into why our coverage went down by -0.08% anyone who knows how to investigate that, lemme know!

yoshokatana:
I think you're one test/branch away from 100% coverage 😄

yoshokatana:
💯 👍

PR #278, opened 2016-06-27
Simplify publishing chain

byronhulcher:
The publishing chain just needs to be an array of publishing functions, not an array of functions that each return a publishing function.


yoshokatana:
👍

PR #281, opened 2016-07-06
log scheduled and published pages

yoshokatana:
* always log when items are scheduled
* always log when pages are published

### DIFF
--- a/lib/services/schedule.js
+++ b/lib/services/schedule.js
@@ -118,7 +118,7 @@ function post(uri, data) {
     ];
 
   return db.batch(ops).then(function () {
-    log('info', 'scheduled', targetUri);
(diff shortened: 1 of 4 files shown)


lib/services/schedule.js
```diff
@@ -121,7 +117,9 @@ function post(uri, data) {
       { type: 'put', key: targetReference, value: JSON.stringify(referencedData) }
     ];
 
-  return db.batch(ops).return(referencedData);
+  return db.batch(ops).then(function () {
+    log('info', 'scheduled', targetUri, `(${data.at})`);
```
yoshokatana: this is a unix timestamp that we can easily grab and see exactly when something has been logged

byronhulcher:
👍

PR #286, opened 2016-07-25
add support for ELK logging

yoshokatana:
* adds support for ELK (elastic, Logstash, Kibana) logging
* set `CLAY_ELK` env variable to enable this
* `CLAY_ELK` should be `domain:port`
* moved `log` into services and exports it, so Clay instances can hook into centralized logging


larrychevres:
thank you, thank you, thank you!

+1

jonwinton:
👍

lib/services/log.js
```diff
@@ -1,19 +1,41 @@
...
+}
+
 // default to console logger, but pretty-print it
 winston.remove(winston.transports.Console);
 winston.add(winston.transports.Console, {
   colorize: true
 });
 winston.default.transports.console.level = logLevel;
 
+addELK(process.env.CLAY_ELK); // CLAY_ELK=domain.com:9200
```
cruzanmo: ELK is now officially called Elastic Stack.

Do we need to prefix with `CLAY_`? Amphora's other environment variables are not prefixed.

yoshokatana: eh, everyone seems to refer to it as ELK still. We can definitely change this in the future. As for prefixing, I want to because hypothetically there could be non-clay logs being generated on the same server and this avoids any collision. (This env variable _specifically_ refers to _Clay_ logs)

PR #287, opened 2016-07-27
Page layout default value

byronhulcher:
Currently in layouts, you are allowed to use strings as keys to layout properties that are meant to reference properties of the page. For instance: "main", "minor", "splashHeader".

Currently, if it has a string value and it can't find a match in page's keys, it will leave the value as is, as a string.  Now, it will be replaced by an empty array.  All layout properties currently are componentLists, so we should replace with an empty list so we can do things like check main.length.  Previously, you would have gotten a false positive if you hadn't defined "main" in page data, as "main".length === 4, so it would evaluate truthily.

### DIFF
--- a/lib/html-composer.test.js
+++ b/lib/html-composer.test.js
@@ -380,7 +380,7 @@ describe(_.startCase(filename), function () {
   describe('refToObj', function () {
     const fn = lib[this.title];
 
-    it('wraps an object with', function () {
+    it('sets data to be the value of an object with key "_ref"', function () {
       const data = { key: 'value'},
         expectedResultData =  { _ref: data },
         newData = fn(data);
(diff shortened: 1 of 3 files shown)


yoshokatana:
I'm ok with this as long as you add a test to bring it back to 100% coverage

cruzanmo:
I agree, needs a test. Otherwise looks good, as `[]` is generally considered empty.

lib/html-composer.test.js
```diff
@@ -374,4 +376,48 @@ describe(_.startCase(filename), function () {
 
     lib(res.req, res, done);
   });
+
+  describe('refToObj', function () {
+    const fn = lib[this.title];
+
+    it('wraps an object with', function () {
```
cruzanmo: Should this be something like "replaces ref with data"?

byronhulcher: good call

cruzanmo:
+1

PR #300, opened 2016-08-22
add available routes to locals

nelsonpecora:
this is useful for things like [url validation](https://trello.com/c/4bXk7e7o/10-validate-custom-urls-kiln) on the client-side, as components like kiln would need a list of the available routes for a site.

it exposes an array of the paths in `res.locals.routes`, e.g. `['/', '/:name']`


gloddy:
👍

PR #301, opened 2016-08-24
switch to hostname from host on redirect

larrychevres:
When using express 'trust proxy' setting req.hostname pulls first from X-Forwarded-Host instead of Host header. If X-Forwarded-Host does not exist, it falls back to Host header (e.g. for local development).


nelsonpecora:
👍

jonwinton:
👍

PR #305, opened 2016-08-31
add available components to locals

nelsonpecora:
adds all available components to locals. this is useful for kiln to know what components are available on a specific clay install, since a layout's _components array is actually just an array of the components that currently exist in that layout.


gloddy:
👍

PR #307, opened 2016-09-12
log errors when component modules error out

nelsonpecora:
this will log errors when component modules error out (and the error isn't just that the module cannot be found)

this fixes most of the issues we've had with eaten logs

_note:_ this won't log when a component module requires another module that doesn't exist (but it will log if that other module exists but throws a different error)

[trello ticket](https://trello.com/c/7mKVZwPI/44-better-silly-logging-when-clay-404-s)


byronhulcher:
👍

PR #312, opened 2016-10-03
added /users route

nelsonpecora:
Two new routes!

* `/users` lists all users in a site, and allows creating new users
* `/users:name` gives data for an individual user, and allows updating and deleting individual users

Notes:

* `:name` is the automatically-generated ID you get when creating a user. Uses the same UID generator as component instance IDs.
* After this I'll start adding oauth stuff that uses the user data

### DIFF
--- a/lib/routes/users.js
+++ b/lib/routes/users.js
@@ -1,0 +1,56 @@
+/**
+ * Controller for Users
+ *
+ * @module
+ */
+
+'use strict';
+
+const _ = require('lodash'),
+  responses = require('../responses'),
+  controller = require('../services/users');
+
+/**
+ * All routes go here.
+ *
+ * They will all have the form (req, res), but never with next()
+ *
+ * @namespace
+ */
+let route = _.bindAll({
+  /**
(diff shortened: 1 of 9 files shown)


lib/routes/users.js
```diff
@@ -0,0 +1,56 @@
...
+  router.post('/', responses.denyReferenceAtRoot);
+  router.post('/', route.post);
+
+  // get and update users
+  router.all('/:name', responses.acceptJSONOnly);
+  router.all('/:name', responses.methodNotAllowed({allow: ['get', 'put', 'delete']}));
+  router.all('/:name', responses.denyTrailingSlashOnId);
+  router.put('/:name', responses.denyReferenceAtRoot);
+  router.get('/:name', responses.getRouteFromDB);
+  router.put('/:name', responses.putRouteFromDB);
```
nelsonpecora: oh haha, there isn't. I can rearrange that

nelsonpecora:
👎  some things in here need to change for auth. I'm gonna roll this up into that PR instead.

PR #315, opened 2016-10-12
add layout ref to state

cruzanmo:
It is useful to have the layout's uri in the DOM as the layout is a component on the page.


gloddy:
Winner!

PR #322, opened 2016-10-26
Return an error message if we have one

byronhulcher:
If we're gonna go to the trouble of creating an error message, the least we could do is make sure it's returned in our response!


nelsonpecora:
👍

PR #323, opened 2016-10-28
authentication

nelsonpecora:
- [trello ticket](https://trello.com/c/GRQw6Oxg/43-oauth-mvp)
- supports apikey, google/twitter/slack oauth, ldap
- adds `/users` (site-specific) keyed by `base65(username @ provider)`
- adds `/auth/<provider>` and `/auth/<provider>/callback` routes
- adds authorization to edit mode, non-get api calls
- adds login page fallback at `/auth/login` which displays enabled providers
- adds logout route at `/auth/logout`
- uses session store for non-apikey auth
- redirects users to previous page when logging in
- pulls display name and image from oauth providers


test/api/users/delete.js
```diff
@@ -0,0 +1,66 @@
+'use strict';
```
gloddy: Shouldn't the filename have `.test` ?

views/clay-logo.svg
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
gloddy: Mind optimizing this SVG?

nelsonpecora: Sure!

gloddy:
Looks healthy to me! 👍

lib/auth.js
```diff
@@ -0,0 +1,446 @@
...
+ * get callback url for a site
+ * note: needs to add/not add initial slash, depending on the site path
+ * @param {object} site
+ * @param {string} provider
+ * @returns {string}
+ */
+function getCallbackUrl(site, provider) {
+  return site.path ?
+    `${references.uriToUrl(site.prefix, null, site.port)}/auth/${provider}/callback` :
+    `${references.uriToUrl(site.prefix, null, site.port)}auth/${provider}/callback`;
```
cruzanmo: url.resolve can do this https://nodejs.org/api/url.html#url_url_resolve_from_to

lib/auth.js
```diff
@@ -0,0 +1,446 @@
...
+function returnUser(user, done) {
+  done(null, user);
+}
+passport.serializeUser(returnUser);
+passport.deserializeUser(returnUser);
+
+/**
+ * create/authenticate against a clay user
+ * @param {object} properties to grab from provider and provider name itself
+ * @param {obejct} site
```
cruzanmo: typo

lib/auth.js
```diff
@@ -0,0 +1,446 @@
...
+        break;
+      case 'google': createGoogleStrategy(site);
+        break;
+      case 'slack': createSlackStrategy(site);
+        break;
+      case 'ldap': createLDAPStrategy(site);
+        break;
+      case 'apikey': createAPIKeyStrategy();
+        break;
+      default: throw new Error(`Unknown provider: ${provider}!`);
```
cruzanmo: Could be useful to doc possible providers / necessary environment vars for each.

lib/auth.js
```diff
@@ -0,0 +1,446 @@
...
+    });
+
+  // add svgs to handlebars
+  _.each(icons, function (icon) {
+    handlebars.registerPartial(icon, compileTemplate(`${icon}.svg`));
+  });
+
+  _.each(providers, module.exports.createStrategy(site)); // allow mocking this in tests
+
+  if (providers.length) {
```
cruzanmo: The above two `_.each` functions could be pulled inside this `if`.

lib/routes/users.js
```diff
@@ -0,0 +1,56 @@
...
+  'post'
+]);
+
+function routes(router) {
+  router.use(responses.varyWithoutExtension({varyBy: ['Accept']}));
+
+  // list users
+  router.all('/', responses.methodNotAllowed({allow: ['get', 'post']}));
+  router.all('/', responses.notAcceptable({accept: ['application/json']}));
+  router.get('/', responses.list());
```
cruzanmo: Do we want to allow this GET if not authenticated?

lib/setup.js
```diff
@@ -11,21 +11,24 @@ bluebird.config({
 });
 
 /**
- * optionally pass in an express app and/or templating engines
+ * optionally pass in an express app, templating engines, and/or providers
+ * note: if no providers are passed in, amphora doesn't protect routes
  * @param {object} [options]
+ * @param {array} [options.providers]
```
nelsonpecora: hmm, they shouldn't be

lib/auth.js
nelsonpecora: neat!

nelsonpecora: Wait, no. `url.resolve` will kill the site path.

lib/auth.js
nelsonpecora: true, done

lib/auth.js
nelsonpecora: I think I'm just gonna exit early if there are no providers 😄

lib/routes/users.js
nelsonpecora: yes

lib/setup.js
cruzanmo: `[] instanceof Array` so `@param {Array}` is correct

lib/auth.js
cruzanmo: ah, you are right

lib/routes/users.js
cruzanmo: I think the list of users should be protected, but I understand arguments on the other side for our use case.

lib/setup.js
nelsonpecora: the official JSDoc documentation seems to do it both ways, hm http://usejsdoc.org/tags-param.html

lib/routes/users.js
byronhulcher: mmmm yeah wait what? Can you elaborate on what's being exposed through this GET? Definitely seems weird to have this publicly available (especailly if it includes providers and usernames?)

lib/auth.js
```diff
@@ -0,0 +1,454 @@
...
+ * @param {object} site
+ * @returns {Promise}
+ */
+function verify(properties, site) {
+  return function (req, token, tokenSecret, profile, done) { // eslint-disable-line
+    var username = _.get(profile, properties.username),
+      imageUrl = _.get(profile, properties.imageUrl),
+      name = _.get(profile, properties.name),
+      provider = properties.provider,
+      uid = site.prefix + '/users/' + users.encode(`${username.toLowerCase()}`, provider);
```
byronhulcher: this will error if `username` is undefined for whatever reason, yeah?

byronhulcher: Also are we able to store non-ASCII characters in keys? just wondering

views/login.handlebars
```diff
@@ -0,0 +1,113 @@
...
+
+      .button:hover {
+        background-color: #fafafa;
+      }
+
+      .button,
+      .button:hover,
+      .button:active,
+      .button:visited {
+        color: #666;
```
byronhulcher: 😈

lib/auth.js
```diff
@@ -0,0 +1,453 @@
...
+  path = require('path');
+
+/**
+ * determine if a route is protected
+ * protected routes are ?edit=true and any method other than GET
+ * @param {object} req
+ * @returns {boolean}
+ */
+function isProtectedRoute(req) {
+  return !!req.query.edit || req.method !== 'GET';
```
cruzanmo: hmm interesting point -- the `edit` query param is part of `clay-kiln` which amphora should not be tied to?

lib/routes/users.js
nelsonpecora: * username
* provider
* display name
* title
* image

lib/auth.js
nelsonpecora: Amphora already knows about edit mode, and (contrary to that old comment in the code) it seems that's the sane place for us to deal with it https://github.com/nymag/amphora/blob/master/lib/html-composer.js#L332

lib/auth.js
```diff
@@ -0,0 +1,453 @@
...
+
+/**
+ * get callback url for a site
+ * note: needs to add/not add initial slash, depending on the site path
+ * @param {object} site
+ * @param {string} provider
+ * @returns {string}
+ */
+function getCallbackUrl(site, provider) {
+  return site.path ?
```
nelsonpecora: all sites have a prefix (it's the host + path), but not all sites have paths

nelsonpecora: let me refactor this to make it more explicit though

lib/auth.js
nelsonpecora: I believe full UTF-8 works for keys, but yagni. if there's no username, it should error out because it means our code can't handle whatever provider they're authenticating against 😄

lib/auth.js
```diff
@@ -0,0 +1,454 @@
...
+/**
+ * middleware to log out. redirects to login page
+ * note: it goes to login page because usually users are navigating from edit mode,
+ * and they can't be redirected back into edit mode without logging in
+ * @param {object} site
+ * @returns {function}
+ */
+function onLogout(site) {
+  return function (req, res) {
+    req.logout();
```
nelsonpecora: it's an method passport adds to the request

lib/auth.js
byronhulcher: Hrm, maybe we should have an explicit check + error it should raise for that?

lib/auth.js
byronhulcher: ah okay! 👍

lib/auth.js
byronhulcher: I'm not a huge fan of truthy values in url params but if we're already treating it truthy its good for me

lib/routes/users.js
byronhulcher: Hrm, seems like kind of a security issue to expose that (here are the vectors you could use to log into our Clay site!)

lib/auth.js
cruzanmo: What is concerning is that we are saying any site built on amphora cannot have an `edit` query param. Would be nice to leave the query params open.

I am ok with this for this PR, but we might want to consider something like a `locals.isProtected` property to indicate that login is required. The sites repo could add `isProtected` to any request that had the `edit` queryparam. That way the individual implementation of clay controls which queryparams are protected.

nelsonpecora: @cruzanmo that's an interesting idea!

lib/auth.js
nelsonpecora: roger

lib/routes.test.js
```diff
@@ -188,17 +189,40 @@ describe(_.startCase(filename), function () {
...
       files.getFiles.returns(['a.b, c.d, e.f']);
 
-      fn(router, {slug: 'example', assetDir: 'example'});
+      fn(router, [], {slug: 'example', assetDir: 'example'});
 
       sinon.assert.calledWith(files.tryRequire, './routes/a');
       sinon.assert.calledWith(files.tryRequire, './routes/a');
     });
 
+    it('inits auth if there are providers', function () {
```
nelsonpecora: it's the routes service making sure it inits auth if it gets passed providers

nelsonpecora:
Per Larry, we're going to allow GETs to users for now, as we'll be keeping everything behind the vpn anyway. In the future (when we start implementing permissions), the specific routes we block will be a configurable thing.

cruzanmo:
+1 -- great upgrade to support auth.

lib/auth.js
```diff
@@ -0,0 +1,470 @@
...
+      case 'ldap': createLDAPStrategy(site);
+        // ldap requires LDAP_URL, LDAP_BIND_DN, LDAP_BIND_CREDENTIALS, LDAP_SEARCH_BASE, and LDAP_SEARCH_FILTER
+        break;
+      case 'apikey': createAPIKeyStrategy();
+        // apikey requires CLAY_ACCESS_KEY
+        break;
+      default: throw new Error(`Unknown provider: ${provider}!`);
+    }
+  };
+}
```
larrychevres: Is it worth it creating a providers folder and just dynamically import provider's strategy creator as well as any other provider-specific settings from there based on provider name?  Simplifies this and moves all those domain-specific logic to their own containers.

nelsonpecora:
We're gonna use this for now, then add the ability to pass in custom passport strategies in the future, per Larry.

PR #327, opened 2016-11-01
Elastic Integration

jonwinton:
Creates a search service in Amphora which can be used to index content. Implementation can be found here: https://github.com/nymag/sites/compare/search-service?expand=1


lib/services/search.js
```diff
@@ -0,0 +1,395 @@
+'use strict';
+
+const es = require('elasticsearch'),
```
gloddy: Different const name perhaps? I see `es` and think `ecmascript`. Not a deal breaker of course.

lib/services/search.js
```diff
@@ -0,0 +1,395 @@
...
+ * @returns {Promise}
+ */
+function initIndex(index, mappings) {
+  return module.exports.client.indices.create({
+    index: index,
+    body: mappings
+  }).then(function () {
+    log('info', 'Successfully created ' + index);
+  })
+    .catch(function (error) {
```
nelsonpecora: indentation

lib/services/search.js
```diff
@@ -0,0 +1,395 @@
...
+}
+
+/**
+ * Indices should have a different name than the alias
+ * Append '_v1' to newly created indices
+ * @param {string} alias
+ * @returns {string}
+ */
+function createIndexName(alias) {
+  return alias + '_v1';
```
nelsonpecora: @dmabuada what are aliases for (and do we still need them)?

lib/services/search.js
```diff
@@ -0,0 +1,395 @@
...
+ * @param {string} index
+ * @returns {Promise}
+ */
+function delIndex(index) {
+  return module.exports.client.indices.delete({
+    index: index
+  }).then(function () {
+    log('info', 'Successfully deleted ' + index);
+  })
+    .catch(function (error) {
```
nelsonpecora: indentation

lib/services/search.js
```diff
@@ -0,0 +1,395 @@
...
+    log('info', 'Elasticsearch cluster is up!');
+    return bluebird.resolve();
+  }).catch(function (error) {
+    log('info', 'Elasticsearch cluster is down!');
+    return bluebird.reject(error);
+  });
+}
+
+/**
+ * Create an Elasticsearch index
```
nelsonpecora: what's the difference between creating an index, creating an alias, and creating a mapping?

lib/services/search.js
```diff
@@ -0,0 +1,395 @@
...
+  return module.exports.client.search({
+    index: index,
+    type: type,
+    body: query
+  }).then(function () {
+
+    // console.log(resp);
+    // let hits = resp.hits.hits;
+
+    // console.log(hits);
```
nelsonpecora: clean this up?

lib/services/search.js
```diff
@@ -0,0 +1,395 @@
...
+function put(index, type, ref, source) {
+  return module.exports.client.index({
+    index: index,
+    type: type,
+    id: ref,
+    body: source
+  }).then(function (resp) {
+    log('info', resp);
+  })
+    .catch(function (error) {
```
nelsonpecora: indentation

lib/services/search.js
```diff
@@ -0,0 +1,395 @@
...
+function del(index, type, ref) {
+  return module.exports.client.delete({
+    index: index,
+    type: type,
+    id: ref
+  }).then(function (resp) {
+    log('info', resp);
+    return resp;
+  })
+    .catch(function (error) {
```
nelsonpecora: indentation

lib/services/search.js
```diff
@@ -0,0 +1,395 @@
...
+function initIndex(index, mappings) {
+  return module.exports.client.indices.create({
+    index: index,
+    body: mappings
+  }).then(function () {
+    log('info', 'Successfully created ' + index);
+  })
+    .catch(function (error) {
+      log('error', error);
+    });
```
nelsonpecora: not rejecting this error?

lib/services/search.js
```diff
@@ -0,0 +1,395 @@
...
+ * @returns {Promise}
+ */
+function batch(ops) {
+  return module.exports.client.bulk({
+    body: ops
+  }).then(function (resp) {
+    let str;
+
+    if (resp && resp.errors === true) {
+      str = ['Client.bulk errored on ' + util.inspect(ops, {depth: 10})];
```
nelsonpecora: why not `let str` here?

lib/services/search.js
```diff
@@ -0,0 +1,395 @@
...
+module.exports.createMappingIfNone = createMappingIfNone;
+module.exports.createIndexName = createIndexName;
+module.exports.convertRedisBatchtoElasticBatch = convertRedisBatchtoElasticBatch;
+module.exports.del = del;
+module.exports.batch = batch;
+module.exports.put = put;
+module.exports.query = query;
+
+// Required for testing to be able to override default client
+module.exports.overrideClient = function (overrideClient) {
```
nelsonpecora: do you need this if you're doing dependency injection in `setup`?

lib/services/search.js
dmabuada: yes we still need aliases. aliases are useful when reindexing and you don't want to lose any data because you can point an index (published-articles_v1) to an alias (published-articles). see here for more info https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-aliases.html

lib/services/search.js
dmabuada: i talked about the diff between aliases + indices elsewhere. mappings are for defining how our data is indexed.

lib/services/search.js
jonwinton: So we need to make this suffix dynamic? Any ideas where that should be passed in? The search service in `sites`?

lib/services/search.js
jonwinton: Nope! Proper indentation per eslint.

lib/services/search.js
jonwinton: Nope! Proper indentation per eslint.

lib/services/search.js
jonwinton: Nope! Proper indentation per eslint.

lib/services/search.js
jonwinton: Nope! Proper indentation per eslint.

lib/services/search.js
larrychevres: In general we don't delete indexes, or at the very least, we leave that to Ops. Do we need to have that? What's the use case.  The whole alias/index relationship is to enable rollback, so when would you delete?

lib/services/search.js
```diff
@@ -0,0 +1,404 @@
...
+  return client.indices.delete({
+    index: index
+  }).then(function () {
+    log('info', 'Successfully deleted ' + index);
+  })
+    .catch(function (error) {
+      log('error', error);
+      return bluebird.reject(error);
+    });
+}
```
larrychevres: In general we don't delete indexes, or at the very least, we leave that to Ops. Do we need to have that? What's the use case. The whole alias/index relationship is to enable rollback, so when would you delete?

jonwinton: Not entirely sure, but shouldn't it be available just in case anyone needed it? I figured that since it's open source someone might want it eventually.

lib/services/search.js
```diff
@@ -0,0 +1,404 @@
...
+module.exports.existsIndex = existsIndex;
+module.exports.initMapping = initMapping;
+module.exports.existsMapping = existsMapping;
+module.exports.createMappingIfNone = createMappingIfNone;
+module.exports.createIndexName = createIndexName;
+module.exports.convertRedisBatchtoElasticBatch = convertRedisBatchtoElasticBatch;
+module.exports.del = del;
+module.exports.batch = batch;
+module.exports.put = put;
+module.exports.query = query;
```
larrychevres: In general, I think I'd like to see assurance that index name and alias can never be the same. If they ever are you could generate errors on some of these ops.  Currently we do that by using simple versioning on index name (e.g. append "_v1"), so I'd like to see a naming standard proposed that is checked before any create/update index operation is called.

lib/services/search.js
jonwinton: @larrychevres I removed the `delIndex` function per our conversation yesterday.

lib/services/search.js
jonwinton: @larrychevres per our conversation yesterday about aliases and our workflow I have made the following updates:

- Changed the [function](https://github.com/nymag/amphora/pull/327/files#diff-0f9d28e9bf5f9c727a7d43583bea7b3fR361)/[variable](https://github.com/nymag/amphora/pull/327/files#diff-0f9d28e9bf5f9c727a7d43583bea7b3fR18) that were named for setting an alias version to now set an index version.
- Added a check [here](https://github.com/nymag/amphora/pull/327/files#diff-0f9d28e9bf5f9c727a7d43583bea7b3fR362) so that a version must to begin with `_v` for consistency 

Per your comment on checking/updating indices, that logic is over in [sites](https://github.com/nymag/sites/blob/search-service/services/search/index.js#L323-L336) and is what we currently do now.

lib/services/search.js
```diff
@@ -0,0 +1,387 @@
...
+ *
+ * @param {string} index
+ * @returns {Promise}
+ */
+function createAliasIfNone(index) {
+  return module.exports.existsAlias(index)
+    .then(function (exists) {
+      if (!exists) {
+        // Indices should have a different name than the alias
+        // Append `aliaSuffix` to newly created indices
```
dmabuada: hmmm? aliaSuffix?

lib/services/search.js
```diff
@@ -0,0 +1,387 @@
...
+module.exports.initIndex = initIndex;
+module.exports.initAlias = initAlias;
+module.exports.createIndexIfNone = createIndexIfNone;
+module.exports.createAliasIfNone = createAliasIfNone;
+module.exports.existsAlias = existsAlias;
+module.exports.existsIndex = existsIndex;
+module.exports.initMapping = initMapping;
+module.exports.existsMapping = existsMapping;
+module.exports.createMappingIfNone = createMappingIfNone;
+module.exports.createIndexName = createIndexName;
```
dmabuada: why is this being exported? and why are we calling it during a batch op in sites? https://github.com/nymag/sites/blob/search-service/services/search/index.js#L267

lib/services/search.js
dmabuada: it shouldn't begin with _v; it should be appended to index names. the name of the mapping file  (e.g. editable-articles.yml) becomes "editable-articles", which becomes the alias. index name is then "editable-articles_v1".

jonwinton: Per conversation with Dalia, here is the function in sites which ensures that index name and alias aren't the same. It adds `_v1` to the mapping's filename: https://github.com/nymag/sites/blob/search-service/services/search/index.js#L323-L336

dmabuada:
This looks like it's good to merge. +1

PR #339, opened 2016-11-28
Elastic Integration (Phase II)

jonwinton:
**This PR Includes Breaking Changes**

New Features:
- Amphora now connects to Elastic Search directly rather than via a service in the implementation of Clay. For reference, [here is the original method](https://github.com/nymag/sites/blob/master/app.js#L67-L82) and this is [the updated method](https://github.com/nymag/sites/blob/search-service/app.js#L66-L69).
- New directory called `elastic` which contains helpers for parsing batch operations from the database. Directory will also house the mappings internal to Amphora (coming soon)
- Added a utility to Amphora that allows you to [dynamically require a module but only if the module exists](https://github.com/nymag/amphora/blob/internal-mappings/lib/utils/module-exists.js). Handy for when you're checking to see if something has been added to a specific Clay implementation but is optional.
- Shifted around and exposed helper methods for parsing batch ops from the dataase.

On The Implementation Side: 
- In any Clay implementation, you can now include a folder called `search` at the root of your instance and implement a more robust system of parsing data for ES. [See example of implementation here](https://github.com/nymag/sites/tree/search-service/search).


nelsonpecora:
Wait, what exactly is the breaking change?

lib/files.js
```diff
@@ -305,9 +334,11 @@ exports.getComponentPackage = control.memoize(getComponentPackage);
 exports.isDirectory = control.memoize(isDirectory);
 exports.tryRequire = control.memoize(tryRequire);
 exports.readFilePromise = control.memoize(readFilePromise);
+exports.getFilesPromise = control.memoize(getFilesPromise);
```
nelsonpecora: what's the difference between this and `readFilePromise`?

jonwinton:
You no longer instantiate the search service in the `app.js` within your Clay instance, all of that is moved into amphora. [This](https://github.com/nymag/sites/blob/search-service/app.js#L66-L69) vs [this]([url](https://github.com/nymag/sites/blob/master/app.js#L67-L82)).

lib/services/search.js
```diff
@@ -358,13 +374,80 @@ function setup(overrideClient) {
...
+
+  return mappings;
+}
+
+// /**
+//  * Create indices that are required for Clay
+//  *
+//  * @return {Promise}
+//  */
+// function createInternalIndices() {
```
nelsonpecora: commented out...?

nelsonpecora:
@jonwinton yeah, but that's not a breaking change in amphora's api, is it? You could still _technically_ init that stuff in your clay instance (though you don't need to anymore)

gloddy:
@jonwinton @nelsonpecora I'd think that's a breaking change to `sites` but not `amphora`.

gloddy:
@jonwinton In the feature list:
`For reference, here is the original method and this is the updated method.`
Both links are the same.

jonwinton:
Ah, sorry about that @gloddy. I've updated the original message.

lib/files.js
jonwinton: `readFilePromise` returns the contents of a file, `getFilesPromise` is like `getFiles` in that it returns all the filenames within a directory. I just needed to have a version of `getFiles` which returned a Promise.

lib/services/search.js
jonwinton: I can remove this from this PR, but this function will be included in the next phase of changes whenever we actually instantiate the indices that are managed by Amphora.

lib/files.js
nelsonpecora: ohhhh makes sense

lib/elastic/db-filters.js
```diff
@@ -0,0 +1,83 @@
...
+ * @returns {boolean}
+ */
+function isOpForPut(op) {
+  return op.type === 'put';
+}
+
+/**
+ * True is component is published ("@published")
+ * @param {object} op
+ * @returns {boolean}
```
cruzanmo: `@param {string} op.key`?

lib/elastic/db-filters.test.js
```diff
@@ -0,0 +1,188 @@
...
+        let op = {
+          type: 'put',
+          key: 'www.nymag.com/scienceofus/pages/cit0k8p6x0000r7reetki2i6k@published',
+          value: '{}'
+        };
+
+        expect(fn(op)).to.equal(true);
+      });
+
+      it('takes an operation and returns false if its key is\'nt a page ref', function () {
```
cruzanmo: `isn\'t` or `is not` -- might want to search replace for this

lib/elastic/db-filters.test.js
```diff
@@ -0,0 +1,188 @@
...
+        let op = {
+          type: 'put',
+          key: 'www.nymag.com/scienceofus/pages/cit0k8p6x0000r7reetki2i6k@published',
+          value: '{}'
+        };
+
+        expect(fn(op)).to.equal(true);
+      });
+
+      it('takes an operation and returns false if its key does\'nt have a published ref', function () {
```
cruzanmo: `doesn\'t` or just `does not`

lib/elastic/helpers.js
```diff
@@ -0,0 +1,297 @@
...
+}
+
+/**
+ * Storing null and undefined for a mapped property
+ * confuses ES (doesn't error, but creates odd behavior).
+ *
+ * @param {string} propertyName
+ * @returns {Function}
+ */
+function removeEmptyValuesForProperty(propertyName) {
```
gloddy: Does this get used anywhere? Popped into my head when I saw this: https://github.com/nymag/amphora/pull/339/files#diff-6612bbbf33f2f172a7fe9e466fbfff28R214

lib/elastic/helpers.js
```diff
@@ -0,0 +1,297 @@
...
+ * and then invoke the function with the value you're
+ * testing as the new argument.
+ *
+ * @param  {String} expected
+ * @return {Function}
+ */
+function compare(expected) {
+  return function (comparison) {
+    return expected === comparison;
+  };
```
cruzanmo: Arrow function could be nice here `return comparison => expected === comparison;`

lib/elastic/helpers.js
```diff
@@ -0,0 +1,297 @@
...
+ * If any of the properties that should be
+ * strings are references, resolve them.
+ *
+ * @param {string} propertyName
+ * @returns {Function}
+ */
+function resolveReferencesForPropertyOfStringType(propertyName) {
+  return function (ops) {
+    return bluebird.all(_.map(ops, function (op) {
+      var result, value = op.value[propertyName];
```
cruzanmo: always will be an op.value object?

lib/elastic/helpers.js
```diff
@@ -0,0 +1,297 @@
...
+ * @param {string} type
+ * @param {Array} ops
+ * @returns {Array}
+ */
+function convertRedisBatchtoElasticBatch(index, type, ops) {
+  let bulkOps = [];
+
+  _.each(ops, function (op) {
+    if (_.isString(op.value)) {
+      op.value = JSON.parse(op.value);
```
cruzanmo: op.value is always an object or JSON string?

lib/services/search.js
```diff
@@ -339,17 +313,59 @@ function batch(ops) {
 }
 
 /**
- * Create the ES Client or an empty object
```
gloddy: Comment not helpful anymore?

lib/services/search.js
```diff
@@ -339,17 +313,59 @@ function batch(ops) {
...
+ * @return {Promise}
  */
 function setup(overrideClient) {
-  if (!module.exports.endpoint && !overrideClient) {
-    throw new Error('No Elastic endpoint or client override');
-  }
+  // Create client + check the connection
+  return createClient(overrideClient)
+    .then(healthCheck)
+    // .then(createInternalIndices)
```
gloddy: Holding this here for the next iteration?

lib/services/search.js
```diff
@@ -339,17 +313,59 @@ function batch(ops) {
 }
 
 /**
- * Create the ES Client or an empty object
+ * [createClient description]
```
gloddy: Delete?

gloddy:
screen shot 2016-11-28 at 3 11 46 pm

lib/services/search.js
```diff
@@ -339,17 +313,59 @@ function batch(ops) {
...
-  // Set the exported client
-  module.exports.client = client = module.exports.endpoint && !overrideClient ? new elastic.Client(_.clone(serverConfig)) : _.assign(module.exports.client, overrideClient);
+/**
+ * Check if the correct indices exist, and if they don't, create them.
+ *
+ * @param {Object} mappings
+ * @returns {Promise}
+ */
+function validateIndices(mappings) {
+  return bluebird.all(_.map(Object.keys(mappings), function (index) {
```
cruzanmo: Could use native map since Object.keys is always an array.

lib/services/search.js
jonwinton: Yep!

lib/services/search.js
```diff
@@ -339,17 +313,59 @@ function batch(ops) {
...
  */
 function setup(overrideClient) {
-  if (!module.exports.endpoint && !overrideClient) {
-    throw new Error('No Elastic endpoint or client override');
-  }
+  // Create client + check the connection
+  return createClient(overrideClient)
+    .then(healthCheck)
+    // .then(createInternalIndices)
+    .then(setupMappings)
```
cruzanmo: Does overrideDir get passed through?

lib/services/search.js
jonwinton: The lodash `_.map` is still faster than native map because it uses a for loop, plus it maps to the patterns that we've already established. I'd prefer to leave it until there's a good performance reason to switch them all over.

lib/elastic/helpers.js
jonwinton: Yessir!

lib/services/search.js
jonwinton: This is fixed, but it's not showing up here..

cruzanmo:
+1

PR #341, opened 2016-12-02
Revert "Elastic Integration (Phase II)"

nelsonpecora:
Reverts nymag/amphora#339

we have an urgent bugfix that needs to go out, and this blocks it. we can re-add afterwards


gloddy:
👍

PR #345, opened 2016-12-08
compose json endpoints

nelsonpecora:
* [trello ticket](https://trello.com/c/z2PMqiq6/73-amphora-should-have-a-composed-json-endpoint)
* `.json` extension endpoints now compose page and component data
* this matches how `.html` endpoints return data
* this works for both GETs and PUTs to the relevant endpoints
* I also cleaned up the `/users` api tests

### DIFF
--- a/lib/routes/components.js
+++ b/lib/routes/components.js
@@ -154,9 +154,8 @@ route = _.bindAll({
       case 'html':
         return this.render(req, res);
       case 'json':
-        return this.getComposed(req, res);
       default:
-        return this.get(req, res);
+        return this.getComposed(req, res);
     }
   },
 
@@ -176,14 +175,13 @@ route = _.bindAll({
             });
         }, res);
       case 'json':
+      default:
         return responses.expectJSON(function () {
           return controller.put(req.uri, req.body, res.locals)
             .then(function () {
               return getComposed(req.uri, res.locals);
             });
         }, res);
-      default:
-        return this.put(req, res);
     }
   },
 
--- a/lib/routes/components.js
+++ b/lib/routes/components.js
@@ -20,6 +20,19 @@ const _ = require('lodash'),
 let validation, route;
 
 /**
+ * get composed component data
+ * @param  {string} uri    of root component
+ * @param  {object} locals
+ * @return {Promise}        composed data of root component and children
+ */
+function getComposed(uri, locals) {
+  return controller.get(uri, locals).then(function (data) {
+    // require here because otherwise it's a circular dependency (via html-composer)
+    return require('../services/composer').resolveComponentReferences(data, locals);
+  });
+}
+
+/**
  * Validation of component routes goes here.
  *
  * They will all have the form (req, res, next).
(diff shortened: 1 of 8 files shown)


nelsonpecora:
Hah, I separated out the default response to extensions and the specific response to the `json` extension. Turns out we actually have code to protect against rando extensions, so that default code's never hit. Updating...

larrychevres:
Hey, a feature that CQ devs used often (and I still use a bit), is a specifier on the endpoint to tell how deep to return.  Might not seem too necessary, but as pages get bigger, and especially when the page is an index or homepage and the data becomes very large, it becomes helpful.

Examples of these:

To just see top level use default .json specifier:
http://author.nymetro.com/content/nymag/daily/entertainment/2016/12/rogue-one-jyn-erso-felicity-jones.json

```
{
    "jcr:createdBy": "ariesman",
    "jcr:created": "Tue Dec 06 2016 17:14:50 GMT-0500",
    "jcr:primaryType": "cq:Page"
}
```

To go 1 level deeper:
http://author.nymetro.com/content/nymag/daily/entertainment/2016/12/rogue-one-jyn-erso-felicity-jones.1.json

```
"jcr:createdBy": "ariesman",
"jcr:created": "Tue Dec 06 2016 17:14:50 GMT-0500",
"jcr:primaryType": "cq:Page",
"jcr:content": {

    "entryType": "Standard",
    "jcr:description": "Explaining <em>Rogue One</em>’s Jyn Erso, Who Is Not Rey",
    "topstorystatus": "Normal",
    "jcr:uuid": "7206dc76-d0bd-4d82-8a54-3df1eb5a3b57",
    "jcr:title": "Explaining Felicity Jones’s <em>Rogue One</em> Character, Jyn Erso, Who Is Not "
  }
}
```

And to fetch the full tree use -1: 
http://author.nymetro.com/content/nymag/daily/entertainment/2016/12/rogue-one-jyn-erso-felicity-jones.-1.json

lib/routes/readme.md
```diff
@@ -103,8 +110,10 @@ module.exports.del = function (ref) {
 ```
 /pages
 /pages/:name
+/pages/:name.json
```
cruzanmo: Looks like we should add `/pages/:name.html`

lib/routes/components.js
```diff
@@ -130,9 +153,9 @@ route = _.bindAll({
     switch (req.params.ext.toLowerCase()) {
       case 'html':
         return this.render(req, res);
-      case 'json': // jshint ignore:line
+      case 'json':
       default:
```
cruzanmo: Should `default` be composed?

lib/routes/readme.md
nelsonpecora: Nice catch!

lib/routes/components.js
nelsonpecora: technically these defaults are never hit, since we [check against a list of approved extensions](https://github.com/nymag/amphora/blob/master/lib/responses.js#L286) before this code is hit

PR #351, opened 2016-12-27
Retrieve mappings passes in a prefix param

dmabuada:
we need to set the elastic prefix var that exists in sites: https://github.com/nymag/sites/blob/f7342b1c873eaae60aa79f4825f33d42ef45d5b7/search/index.js#L41:L62


jonwinton:
Should this not stay a part of Amphora as a helper method but allow a prefix to be passed into the function as well? I feel like this logic shouldn't be the responsibility of a Clay implementation to write since the rest of the Elastic functionality is handled by Amphora.

PR #352, opened 2016-12-28
Handle sites that have subdirectories

dmabuada:


lib/services/sites.js
```diff
@@ -97,9 +97,13 @@ function getSite(host, path) {
  * @returns {object}
  */
 function getSiteFromPrefix(prefix) {
-  const split = prefix.split('/');
+  const split = prefix.split('/', 3),
+    site = sites.getSite(split[0], '/' + split[1]);
```
jonwinton: This just needs to call `getSite`

lib/services/sites.js
```diff
@@ -97,9 +97,13 @@ function getSite(host, path) {
...
 
-  return getSite(split[0], '/' + split[1]);
+  // handle sites that have a "subdirectory" path, e.g. /daily/intelligencer
+  if (typeof site === 'undefined') {
+    return sites.getSite(split[0], '/' + split[1] + '/' + split[2]);
```
jonwinton: Same as above

PR #357, opened 2017-01-05
Search Config

jonwinton:
Creates a config object that allows us to pass data to the search service from a Clay implementation.

### DIFF
--- a/lib/setup.js
+++ b/lib/setup.js
@@ -23,13 +23,14 @@ bluebird.config({
  * @returns {Promise}
  */
 module.exports = function (options) {
-  let app, engines, providers, router, sessionStore;
+  let app, engines, providers, router, sessionStore, searchConfig;
 
   options = options || {};
   app = options.app || express(); // use new express app if not passed in
   engines = options.engines; // will be undefined if not passed in
   providers = options.providers || [];
   sessionStore = options.sessionStore;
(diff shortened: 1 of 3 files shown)


lib/setup.js
```diff
@@ -23,13 +23,14 @@ bluebird.config({
...
 module.exports = function (options) {
-  let app, engines, providers, router, sessionStore;
+  let app, engines, providers, router, sessionStore, searchConfig;
 
   options = options || {};
   app = options.app || express(); // use new express app if not passed in
   engines = options.engines; // will be undefined if not passed in
   providers = options.providers || [];
   sessionStore = options.sessionStore;
+  searchConfig = options.search || {}; // Config object to pass to the search setup
```
dmabuada: lol so basically just passing in config instead of the search engine

dmabuada: also why does this default to an empty object?

jonwinton: Because that's the default we'll want to pass to the search service for now. We'll eventually make it so if it doesn't exist the search service doesn't instantiate

dmabuada:
+1..but stil not 100% happy that we're basically just passing in a string to Amphora. At some point, we should pass in the search client itself so people can add/remove it from their app

jonwinton:
Agreed, we're just moving in stages right now.

PR #363, opened 2017-01-18
preload schemas

nelsonpecora:
* pass schemas to `_componentSchemas` state property so they can be used by templates
* allows [kiln schema preloading](https://github.com/nymag/clay-kiln/pull/736)

### DIFF
--- a/lib/html-composer.js
+++ b/lib/html-composer.js
@@ -99,7 +99,7 @@ function applyOptions(options, res) {
     _.set(state, '_componentSchemas', _.map(componentList, function (name) {
       return {
         name: name,
-        schema: schema.getSchema(files.getComponentPath(name))
+        schema: schema.getSchemaPath(files.getComponentPath(name))
       };
     }));
 
--- a/lib/html-composer.js
+++ b/lib/html-composer.js
@@ -15,6 +15,8 @@ const _ = require('lodash'),
   multiplexTemplates = require('multiplex-templates'),
   references = require('./services/references'),
   composer = require('./services/composer'),
+  schema = require('./schema'),
+  files = require('./files'),
   media = require('./media'),
   url = require('url'),
   defaultHtmlTemplateName = 'template';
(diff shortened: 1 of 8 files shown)


lib/html-composer.js
```diff
@@ -94,6 +96,12 @@ function applyOptions(options, res) {
 
     // can't assign to result, or we'd get a circular reference in refs.
     _.set(state, '_components', componentList);
+    _.set(state, '_componentSchemas', _.map(componentList, function (name) {
+      return {
+        name: name,
+        schema: schema.getSchema(files.getComponentPath(name))
+      };
+    }));
```
cruzanmo: Is kiln the only component interested in this data? I am wondering how this will affect page-load performance, as it increases the size of the state object. Perhaps there could be another end-point to return all the schemas which kiln could call.

nelsonpecora: Hmm. It does increase the size of the state object, but not by a super huge amount. Alternatively, we could give a list of filepaths rather than loading the files, but that would require some odd kiln workarounds.

lib/schema.js
```diff
@@ -18,4 +19,15 @@ function getSchema(dir) {
   return schema;
 }
 
+function getSchemaPath(dir) {
+  if (fs.existsSync(path.join(dir, 'schema.yml'))) {
+    return path.join(dir, 'schema.yml');
+  } else if (fs.existsSync(path.join(dir, 'schema.yaml'))) {
+    return path.join(dir, 'schema.yaml');
```
cruzanmo: Minor possible tweak: 'schema.yml' and 'schema.yaml' could be const at top.

lib/html-composer.js
```diff
@@ -94,6 +96,12 @@ function applyOptions(options, res) {
 
     // can't assign to result, or we'd get a circular reference in refs.
     _.set(state, '_components', componentList);
+    _.set(state, '_componentSchemas', _.map(componentList, function (name) {
```
cruzanmo: `_componentSchemaPaths`?

lib/schema.js
nelsonpecora: Good idea

lib/html-composer.js
nelsonpecora: Eh, I think this is better because they're not just an array of path strings (but rather objects with some properties)

PR #367, opened 2017-01-26
update unpub notify to send url

larrychevres:
The unpublish notifier was sending the url of the internal clay page, which is of no use to external systems trying to handle unpublish of a public url.  This update sends the public url of the unpublished page (in json package).


nelsonpecora:
👍

PR #369, opened 2017-01-27
update memory leak limit to appropriate level

larrychevres:
Helps make the logs more readable. Stops reporting false memory leak conditions.


dmabuada:
+1

PR #377, opened 2017-02-13
Removes all sitemap logic from Amphora

cperryk:
As part of spinning the Cut off to its own domain and meeting its new sitemap requirements ([Trello](https://trello.com/c/deJDnM1D/25-thecut-com-articles-xml-sitemap)), we're stepping away from the old [utilities system](https://github.com/nymag/utilities/blob/0ba342e805b34495ad5e047d7b50470c9251846b/sitemap/run.py) that currently generates our sitemaps. A previous effort to modernize sitemap generation involved adding it to Amphora. While the functionality is there, we're not actually using it, and because Amphora already does a lot, we now want to remove sitemap generation and put it elsewhere (e.g. in `sites` itself). This PR removes sitemap generation from Amphora.


lib/services/db.test.js
```diff
@@ -156,6 +156,23 @@ describe(_.startCase(filename), function () {
...
+          const pipe = fn({
+            json: false
+          });
+
+          pipe.on('data', onData);
+          pipe.on('end', onEnd);
+          pipe.on('error', onError);
+        });
+    });
   });
```
cperryk: Note: This new test is needed because without sitemap tests istanbul will  complain about `db.list` never running with `json` set to `false`.

lib/services/components.js
```diff
@@ -446,7 +446,7 @@ function getPossibleTemplates(name, templateName) {
 /**
  * Get the first template found for a particular component
  * @param {string} name  i.e., 'article', 'paragraph', 'content', 'my-layout'
- * @param {string} templateName  i.e., 'template', 'sitemap', 'apple', 'facebook'
+ * @param {string} templateName  i.e., 'template', 'apple', 'facebook'
```
jasondecastro: The one on line 433 has brackets around `templateName` but this one doesn't, what's the difference?

cperryk: Dunno. The other probably shouldn't have been in brackets in the first place because that var can't be an array. Our commenting around functions is kinda all over the place. I've removed the brackets in a new commit.

PR #378, opened 2017-02-14
Moves page composer function to composer service

cperryk:
We're moving sitemap generation logic out of Amphora ([Trello](https://trello.com/c/deJDnM1D/25-thecut-com-articles-xml-sitemap)). Part of the logic involves composing page objects. This PR move the function that composes pages into the `composer` service, thereby exposing it to modules outside of Amphora.

This change also aligns `routes/pages.js` more closely with `routes/components.js`; both now are responsible only for `get`-ing the proper object; the `composer` owns composing.

### DIFF
--- a/lib/routes/pages.js
+++ b/lib/routes/pages.js
@@ -11,8 +11,6 @@ const _ = require('lodash'),
   htmlComposer = require('../html-composer'),
   controller = require('../services/pages'),
   db = require('../services/db'),
-  components = require('../services/components'),
-  references = require('../services/references'),
   acceptedExtensions = {
     html: 'text/html',
     json: 'application/json'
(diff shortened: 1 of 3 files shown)


jonwinton:
I have questions about why we're removing the site map fuctionality from amphora. Are we saying that anyone who uses amphora has to build their own site map generator? I might have missed some bigger conversations but a little context would help.

cperryk:
@jonwinton Sure. As you know, lately we've talked about trimming Amphora down to its essential parts, offloading some of its logic and opinions to microservices and plugins. @larrychevres thinks sitemap generation should be considered outside Amphora's scope, and I agree. While the sitemap generation process depends on Amphora's various services, none of those services or other parts of Amphora depend on sitemaps. The sitemap logic that was already in Amphora was very rigid, and NyMag itself needs it to be considerably more flexible because sitemap rules, both about which pages appear in a sitemap and how their content is represented, will vary across both sites and sitemap types.

What I'm currently working on is building a lib that will provides a simple sitemap middleware that basically replicates the old functionality that we just removed from Amphora, but also provides tools for building your own sitemap generation logic. So, no, Clay devs will not have to build their own sitemap generator. They would be able to do something like this:

```
const sitemaps = require('amphora-sitemaps');
app.use(sitemaps.pageSitemap(someAmphoraInstance));
```

PR #380, opened 2017-03-02
Adds lastModified to page object

cperryk:
[Trello](https://trello.com/c/deJDnM1D/25-thecut-com-articles-xml-sitemap)

This change adds a `lastModified` property to the page object and updates Amphora's tests accordingly.

This data is needed for sitemaps — by our requirements, each URL in a sitemap must have a `lastmod` element.

### DIFF
--- a/lib/services/pages.js
+++ b/lib/services/pages.js
@@ -282,6 +282,8 @@ function publish(uri, data, locals) {
         // make public
         addOpToMakePublic(ops, prefix, publicUrl, uri);
 
+        pageData.lastModified = new Date().toISOString();
+
         // add page PUT operation last
         return addOp(references.replaceVersion(uri, published), pageData, ops);
       });
--- a/lib/services/pages.test.js
+++ b/lib/services/pages.test.js
@@ -16,7 +16,8 @@ const _ = require('lodash'),
 describe(_.startCase(filename), function () {
   const timeoutConstant = 100;
   let sandbox,
-    savedTimeoutConstant;
+    savedTimeoutConstant,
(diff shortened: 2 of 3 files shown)


lib/services/pages.test.js
```diff
@@ -26,6 +27,7 @@ describe(_.startCase(filename), function () {
     sandbox.stub(notifications, 'notify');
     sandbox.stub(timer);
     sandbox.stub(winston);
+    clock = sinon.useFakeTimers();
```
cruzanmo: Can use `sandbox.useFakeTimers();` so that it does not need to be manually restored.

lib/services/pages.js
```diff
@@ -282,6 +282,8 @@ function publish(uri, data, locals) {
         // make public
         addOpToMakePublic(ops, prefix, publicUrl, uri);
 
+        pageData.lastModified = new Date().toISOString();
```
cruzanmo: Do we want this in amphora? The publishing chain could handle it in the sites instance.

cperryk: Not sure. I see two arguments for putting it here:

*  Amphora plugins, like `clay-sitemaps`, can expect it to be there.
* Either way, Amphora must make an exception for the `lastModified` key in `references.omitPageConfiguration`. Amphora hard-coding an exception like this because of logic outside of it is bad architecture, IMO. (This is what's going on with `urlHistory`).

cperryk:
Closing in favor of [this sites ticket](https://github.com/nymag/sites/pull/2689).

PR #381, opened 2017-03-15
Using the Promise.defer polyfill util

jonwinton:
Migrating `bluebird.defer` to eliminate a warning in the logs

http://bluebirdjs.com/docs/api/deferred-migration.html


nelsonpecora:
👍  but we should fix it for reals → https://github.com/nymag/amphora/issues/382

Issue #386, opened 2017-04-14
Amphora Plugins


jonwinton:
This has come up regarding features like the 'page-list' in Kiln. Amphora should not be responsible for updating an Elastic index, but we do want to easily hook into different stages of the Amphora lifecycle to keep external services in sync. 

Plugins must be registered in the app at Amphora instantiation (1) and expose hooks that Amphora can call (2).

Hooks:
- init
- save
- delete
- ???

jonwinton:
First pass at implementation: https://github.com/nymag/amphora/commit/b33974c2d78e31851dfc5d5780f32dfea42d59cf

Needs tests and comments

jonwinton:
Some plugins may want to query the database to resolve data for components, how should they do this? Should Amphora pass along a reference to the DB service to plugins at initialization? Should the plugins be required to connect to the DB on their own?

amycheng:
I think Amphora should pass along a reference to the DB service. What's happening with `sites/services/db`? Would that be deprecated with plugins or has it already been deprecated because of the pubsub stuff and the new `model.js` conventions (ie avoiding using the db service)?

I also think that its a better separation of concerns if app/plugins avoid interacting with the DB directly; all of that should be delegated to Amphora.

cperryk:
The [clay-sitemaps-generator](https://github.com/nymag/clay-sitemaps) I built is basically an Amphora plugin. Although we'll probably refactor this to rely less on Amphora and more on ElasticSearch at some point, it gives us some sense of what a plugin might need to do. Here's some of the stuff it does:

* Stream page objects from the database matching a prefix.
* Composes (or "explodes" or whatever) pages.
* Renders pages.
* Renders component instances.
* Transforms component/page URIs into other things -- e.g. into component names, instance names, swap versions, etc.

I agree with @amycheng -- I suspect reading Amphora data would be a pretty common task and it shouldn't be the concern of a plugin to implement a way to do that.

Somewhat tangential, but, on that last bullet point: We have many "turn page/cmpt URI into something else" utility functions across our systems ([here](https://github.com/nymag/sites/blob/master/services/db.js) [here](https://github.com/nymag/sites/blob/master/global/kiln-js/references.js) [here](https://github.com/nymag/clay-kiln/blob/master/services/references.js) [here](https://github.com/nymag/homepage-bot/blob/master/lib/components.js)). We should build a `clay-references` package that Amphora and everything else uses to do this kind of work.

jonwinton:
[...]

PR #395, opened 2017-04-21
Pass along page uri in headers

dmabuada:
https://trello.com/c/ElrhLZue/67-add-pageuri-to-request-headers

### DIFF
--- a/lib/services/notifications.js
+++ b/lib/services/notifications.js
@@ -12,9 +12,10 @@ const _ = require('lodash'),
  * @param {object} data
  * @returns {Promise}
  */
-function callWebhook(event, url, data) {
+function callWebhook(event, url, data, uri) {
   const headers = {
-      'X-Event': event
+      'X-Event': event,
+      'page-uri': uri
     },
     options = {
       method: 'POST',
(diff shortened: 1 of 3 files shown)


lib/services/notifications.js
```diff
@@ -12,9 +12,10 @@ const _ = require('lodash'),
  * @param {object} data
  * @returns {Promise}
  */
-function callWebhook(event, url, data) {
+function callWebhook(event, url, data, uri) {
   const headers = {
-      'X-Event': event
+      'X-Event': event,
+      'page-uri': uri
```
cruzanmo: Nice, I generally like this approach as it allows the JSON to remain untouched and allows webhooks to know the URI of the resource triggering the event.

I am wondering if we should use something like `Link: uri + ';rel=meta'`. The `Link` header has be proposed by IETF in 2010, although I do not know how widely it is used: https://www.rfc-editor.org/rfc/rfc5988.txt and https://www.w3.org/wiki/LinkHeader

One benefit of `Link` is that it is more generic and would cover the use-case of not only page publish events but also component instance publish events should we want to support them in the future.

dmabuada: The reason I shied away from Link is because it looks like it's mostly used for pagination purposes? But I don't see why we couldn't use it for this. I agree that we should generalize it more, though.

jonwinton:
I have the following issues, but I don't think I can be a blocker for this.

1. I don't like the idea of passing along (another) a non-standard header in the POST. If the page uri is needed, it should be part of the body of the POST
2. We expose the `db.batch` event that anything can tap into and has access to every piece of data that passes through the DB. I don't see why we don't use that.
3. It is my believe that we should ONLY send the page uri in the POST body and not the full JSON payload. If a service requires data for that page then it can do it's own request to the page uri OR just be built off the DB event/[soon to be implemented plugins](https://github.com/nymag/amphora/issues/386)

PR #401, opened 2017-04-26
GetSiteFromPrefix Update

jonwinton:
The current `getSiteFromPrefix` function only accommodates for sites with one subdirectory in the `path` of a site. This updates makes it so that the function will iterate through a uri until it finds a match or return undefined.

### DIFF
--- a/lib/services/sites.test.js
+++ b/lib/services/sites.test.js
@@ -28,6 +28,15 @@ describe(_.startCase(filename), function () {
         prefix: 'd/e',
         assetDir: 'public',
         assetPath: '/e'
+      },
+      c: {
+        dir: 'z/c',
+        slug: 'c',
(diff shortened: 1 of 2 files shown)


lib/services/sites.test.js
```diff
@@ -28,6 +28,15 @@ describe(_.startCase(filename), function () {
         prefix: 'd/e',
         assetDir: 'public',
         assetPath: '/e'
+      },
+      c: {
+        dir: 'z/c',
+        slug: 'c',
+        host: 'h',
+        path: '/i/j',
```
dmabuada: does this test always assume a path exists? can we write a test for root sites that don't have paths?

dmabuada: ok i see it now.

PR #402, opened 2017-04-26
Amphora Plugins

jonwinton:
Implements [plugins](https://github.com/nymag/amphora/issues/386) in Amphora!

Hooks:
- init
- save
- delete
- routes

Plugins are passed into Amphora at initialization as an array of modules that export any of three functions that are named above:

```
return amphora({
      app: app,
      engines: templateEngines,
      providers: ['apikey', amphoraProvider],
      sessionStore: redisStore,
      plugins: [
        amphoraSearch
      ]
});
```

For the `routes` hook an instance of an Express Router is passed to the hook that is specific _for a specific site_. This allows you to add routes to each site in your Clay instance.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -2,7 +2,7 @@ machine:
   timezone:
     America/New_York
   node:
-    version: 6.0
+    version: 6.10.0
 test:
   post:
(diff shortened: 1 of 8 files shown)


nelsonpecora:
👍🏼

PR #421, opened 2017-07-12
Environment Variables

jonwinton:
Per [ticket](https://trello.com/c/amgms8v6/19-exposing-env-variables-to-client).

Register the needed environment variables with Amphora at instantiation time:
```javascript
   amphora({
      app: app,
      renderers: {
        html: amphoraHtml,
        default: 'html'
      },
      env: ['FOO', 'BAR']
    }).then(....)
```

These will then be resolved into an object when a page is requested for render _only in edit mode_. They will be sent in the `_envVars` property on the `state` object.

### DIFF
--- a/lib/render.js
+++ b/lib/render.js
@@ -1,6 +1,6 @@
 'use strict';
 
-var uriRoutes, renderers, envVars;
+var uriRoutes, renderers;
 const _ = require('lodash'),
   media = require('./media'),
   schema = require('./schema'),
@@ -44,7 +44,6 @@ function registerRenderers(renderObj) {
  * @param  {Array} envArray
  */
 function registerEnv(envArray) {
-  envVars = envArray;
   // Export the renderers object
   module.exports.envVars = envArray;
 }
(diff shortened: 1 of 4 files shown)


mheiber:
trust no one with environment variables

lib/render.js
```diff
@@ -66,6 +77,22 @@ function transfer(from, to, fromProp, toProp) {
...
 /**
+ * Look through env variables array and return
+ * the variables from the `process.env` bject
+ *
+ * @return {Object}
+ */
+function resolveEnvVars() {
+  var obj = {};
+
+  _.map(module.exports.envVars, function (val) {
```
mheiber: `map` is for transforming an array, but this just side-effects. You could use `Array.prototype.forEach`, `_.each` or some other `_` method to better-communicate intent

lib/render.js
```diff
@@ -38,6 +39,16 @@ function registerRenderers(renderObj) {
 }
 
 /**
+ * Register env variables to be referenced
+ *
+ * @param  {Array} envArray
+ */
+function registerEnv(envArray) {
```
mheiber: Is this API documented?

PR #422, opened 2017-07-17
`@published` Specific Lists

jonwinton:
Discovered a feature that would be really handy syncing data between services and re-indexing: list out only `@published` instances of components and pages.

This PR adds the ability to pass `?published=true` to `/pages` and `/components/:name/instances` to list only their published versions.

**Pages Example**
https://dl.dropboxusercontent.com/s/c4m98dmbpcccjfg/AAE691F7-2E76-4F5C-95F6-F6DE6BFF48B7-1332-00004F5678823CFD.gif?dl=0

**Components Example**
https://dl.dropboxusercontent.com/s/jac42rujs6bqi4i/F197D0D3-E5A1-482A-889C-B3A2A5DBED4E-1332-00004EC32B18BB29.gif?dl=0


jonwinton:
Or what about `/pages@published` and `/components/foo/instances@published`? 

Could be more consistent with other patterns and query params are always kind of weird. Would elevate this pattern to a first-class citizen within routing and not bury the flag.

gloddy:
@jonwinton I like it. Matches the pattern in the store perfectly, and avoids growing a query param collection.

nelsonpecora:
I agree with Christian

lib/responses.js
```diff
@@ -476,6 +489,34 @@ function listWithoutVersions() {
...
+  const publishedString = '@published',
+    options = {
+      transforms() {
+        return [filter({ wantStrings: true }, function (str) {
+          return str.indexOf(publishedString) !== -1;
+        })];
+      }
+    };
+
+  if (!req.isAuthenticated()) {
```
nelsonpecora: why are these GETs behind authentication, but other page stuff isn't?

PR #424, opened 2017-07-19
Permissions

jonwinton:
_Updated 9/2_

First iteration of permissions in Amphora. 

**Two permissions levels:** 
- Admin
- Write

**Implementation Details**
- API Keys are assumed to be `admin` level
- Users need a `auth` property set, otherwise errors get thrown
- `401` response method added
- Authentication levels set on each route for `PUT`, `DELETE` and `POST`. At a minimum, `write` level access should be added
- a `users/all-sites` route added for creating/modifying/removing new users for each site with one request (`PUT`, `POST`, `DELETE)

**Further Changes Necessary**
- [ ] Test to 100% coverage once Kiln needs are confirmed

### DIFF
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -11,7 +11,12 @@ const _ = require('lodash'),
   basicAuth = require('basic-auth'),
   fs = require('fs'),
   path = require('path'),
-  USER_LEVELS = {
+  AUTH_LEVELS_MAP = {
+    ADMIN: 'admin',
+    WRITE: 'write',
+    READ: 'read'
+  },
+  AUTH_LEVEL_VALUES = {
     admin: 0,
     write: 1,
     read: 2
@@ -26,13 +31,13 @@ const _ = require('lodash'),
  * @return {Boolean}
  */
 function checkAuthLevel(userLevel, requiredLevel) {
-  console.log(userLevel);
   // User has to have an auth level set
   if (!userLevel) {
     throw new Error('User does not have an authentication level set');
   }
+
   // Compare levels, the lover you go the more permissions you have
-  return USER_LEVELS[userLevel] <= USER_LEVELS[requiredLevel];
+  return AUTH_LEVEL_VALUES[userLevel] <= AUTH_LEVEL_VALUES[requiredLevel];
 }
 
 /**
@@ -45,11 +50,8 @@ function checkAuthLevel(userLevel, requiredLevel) {
  */
 function withAuthLevel(requiredLevel) {
   return function (req, res, next) {
-    if (!req.user) {
-      // TODO: dssdsdfdsf
-      // If no user is set, we might be dealing with an API key, let the request proceed
-      next();
-    } else if (checkAuthLevel(_.get(req, 'user.auth', ''), requiredLevel)) {
+    console.log(req.user);
+    if (checkAuthLevel(_.get(req, 'user.auth', ''), requiredLevel)) {
       // If the user exists and meets the level requirement, let the request proceed
       next();
     } else {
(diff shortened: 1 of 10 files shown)


lib/auth.js
```diff
@@ -6,10 +6,59 @@ const _ = require('lodash'),
...
+ * Get the user auth level and check it against the
+ * required auth level for a route. Send an error
+ * if the user doesn't have permissions
+ *
+ * @param  {String} requiredLevel
+ * @return {Function}
+ */
+function withAuthLevel(requiredLevel) {
+  return function (req, res, next) {
+    if (checkAuthLevel(_.get(req, 'user.auth', ''), requiredLevel)) {
```
mheiber: Why not just pass `user.auth` instead of defaulting to empty string and checking for falsiness in `checkAuthLevel`? Seems simpler and less error-prone.

lib/auth.js
```diff
@@ -6,10 +6,59 @@ const _ = require('lodash'),
...
+
+/**
+ * Get the user auth level and check it against the
+ * required auth level for a route. Send an error
+ * if the user doesn't have permissions
+ *
+ * @param  {String} requiredLevel
+ * @return {Function}
+ */
+function withAuthLevel(requiredLevel) {
```
mheiber: I like that this is middleware that takes an argument, really useful

lib/routes/components.js
```diff
@@ -233,6 +234,7 @@ function routes(router) {
   router.all('/:name@:version', responses.denyTrailingSlashOnId);
   router.get('/:name@:version', route.get);
   router.put('/:name@:version', responses.denyReferenceAtRoot);
+  router.put('/:name@:version', withAuthLevel(authLevels.WRITE));
```
mheiber: very easy to read

lib/routes/users.js
```diff
@@ -49,7 +51,9 @@ function routes(router) {
   router.all('/:name', responses.denyTrailingSlashOnId);
   router.get('/:name', responses.getRouteFromDB);
   router.put('/:name', responses.denyReferenceAtRoot);
+  router.put('/:name', withAuthLevel(authLevels.ADMIN));
```
mheiber: I think that when we POST the same user info multiple times, *BOTH* users end up existing. This hasn't been a problem in practice, but will become one with the new auth system.
Can you add checking-for-dupes to the user creation route handler?

lib/auth.js
```diff
@@ -6,10 +6,59 @@ const _ = require('lodash'),
...
   fs = require('fs'),
-  path = require('path');
+  path = require('path'),
+  AUTH_LEVELS_MAP = {
+    ADMIN: 'admin',
+    WRITE: 'write',
+    READ: 'read'
+  },
+  AUTH_LEVEL_VALUES = {
+    admin: 0,
```
mheiber: where my bit mask at ?

lib/auth.js
```diff
@@ -6,10 +6,59 @@ const _ = require('lodash'),
...
+    write: 1,
+    read: 2
```
mheiber: looks good for now. If we need permissions that are not strictly-ordered in future it would be fun to use bit masks

lib/routes/users.js
```diff
@@ -49,7 +51,9 @@ function routes(router) {
   router.all('/:name', responses.denyTrailingSlashOnId);
   router.get('/:name', responses.getRouteFromDB);
```
nelsonpecora: should `GET /users/:key` be protected as well?

nelsonpecora:
Youuuuuu should update the PR's description 😉

lib/auth.js
jonwinton: If `user` is undefined then an error will be thrown about not being able to read `auth` of undefined. Lodash's get method lets us choose a value to pass in in case the desired property doesn't exist, so this is just insurance that we get to the `checkAuthLevel` function.

PR #426, opened 2017-07-22
Request Route Params

jonwinton:
Adds the route params from `req.params` to the `locals` object.

Use case: pages who have one template for multiple routes which varies content based on the param value (i.e. section pages).

### DIFF
--- a/lib/render.js
+++ b/lib/render.js
@@ -222,8 +222,8 @@ function renderPage(uri, req, res) {
 
   // Add request route params and request queries
   // to the locals object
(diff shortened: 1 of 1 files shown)


mheiber:
@jonwinton seems like a nice idea, I can't think of any downside, is there a downside?

mheiber:
Good idea, will be nice to have access to the req params. Do we have access to req.query already? That would be nice to have, too.

nelsonpecora:
locals.query should already be in there, yeah?

PR #427, opened 2017-07-26
add plugin publish hook

cruzanmo:
This adds a `publish` hook for plugins. On publish both the `save` and `publish` hooks are executed. In writing tests, I noticed that the `batch` method provides different args to the hook than the `put` and `del` methods; perhaps we should execute `delete`, `save`, `publish` for each batch op so that the args that the plugin hooks receive are consistent and so that we can remove the assumption that all ops in the batch are the same type. Or we could map `put` and `del` methods to mirror batches, e.g. `put('key', 'val')` => `executeHook('save', [{type: 'put', key: 'key', value: 'val}])`


lib/services/db.js
```diff
@@ -215,6 +216,91 @@ function formatBatchOperations(ops) {
...
+ *
+ * @param {object} hookOps
+ * @param {Array} hookOps.put
+ * @param {Array} hookOps.del
+ * @param {object} op
+ * @returns {object}
+ */
+function addToHookOps(hookOps, op) {
+  if (hookOps[op.type]) {
+    hookOps[op.type].push(_.pick(op, 'type', 'key', 'value'));
```
jonwinton: Two questions?

1. Any reason to not just `return hookOps.concat(_.pick(op, 'type', 'key', 'value'))`
2. Why use `_.pick` here? The API for the an op is consistent, unless there's something I'm not thinking of.

lib/services/db.js
```diff
@@ -215,6 +216,91 @@ function formatBatchOperations(ops) {
...
+function triggerPlugins(method, args) {
+  const hookOps = dbOpsToHookOps(method, args),
+    // assumes page publish batches always have page put as the last op
+    lastPutKey = _.get(_.last(hookOps.put), 'key'),
+    isPublishedPageBatch = method === 'batch' && hookOps.put.length > 1 &&  _.includes(lastPutKey, '/pages/') && _.endsWith(lastPutKey, publishedVersionSuffix);
+
+  if (hookOps.put.length) {
+    plugins.executeHook('save', hookOps.put);
+    if (isPublishedPageBatch) {
+      plugins.executeHook('publish', {
```
jonwinton: Is there a use case for not wanting to just process a `save` of published data as a `publish` hook? This would give a handler two chances to run on one save of data and I don't know if that's a little too flexible. Thoughts?

lib/services/db.js
cruzanmo: Good catch, _.pick was left over from when I was removing the `type` property, but now _.pick is no longer needed.

concat creates a new Array, which is not necessary / is not as performant.

lib/services/db.js
cruzanmo: Yes that is true. I think the normal case would be for a plugin to use either `save` or `publish`. 

I was defining `save` as any instance being saved, and `publish` as the data saved to @publish instances during a page publish batch.

We could have the definition be something like, `save` is when anything is saved while editing. `publish` is when a user clicks the publish button. Does that sound better?

lib/services/db.js
jonwinton: 👍  I'm good with both of those answers

lib/services/db.js
jonwinton: After speaking to Morgan via Slack I'm of the opinion that we should call both and handlers can decide which to use. You might want to index both editable and published data for some reason and `save` is your catch for that.

lib/services/db.js
```diff
@@ -215,6 +216,91 @@ function formatBatchOperations(ops) {
...
+/**
+ *
+ * @param {string} method
+ * @param {Array} args
+ */
+function triggerPlugins(method, args) {
+  const hookOps = dbOpsToHookOps(method, args),
+    // assumes page publish batches always have page put as the last op
+    lastPutKey = _.get(_.last(hookOps.put), 'key'),
+    isPublishedPageBatch = method === 'batch' && hookOps.put.length > 1 &&  _.includes(lastPutKey, '/pages/') && _.endsWith(lastPutKey, publishedVersionSuffix);
```
jonwinton: Hmmm what happens if you do a `PUT` to a single `@published` instance of a component, it wouldn't be caught by this, correct?

upgrade.js files can update published instances of any component, shouldn't allow handlers to get those changes too? I would double-check that single `PUT's` trickled to the `publish` hook

cruzanmo: correct, a single put op would return false due to `hookOps.put.length > 1`

lib/services/db.js
nelsonpecora: Publishing also happens when scheduled items are triggered, not just from immediate user input 😉

jonwinton: Tested scheduling with these changes and it worked perfectly 👍

PR #431, opened 2017-08-30
Auth Url Patch

jonwinton:
Making sure we get the right site when re-directing to login.

### DIFF
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -7,6 +7,7 @@ const _ = require('lodash'),
   session = require('express-session'),
   flash = require('express-flash'),
   handlebars = require('handlebars'),
+  sites = require('./services/sites'),
   basicAuth = require('basic-auth'),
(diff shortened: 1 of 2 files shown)


lib/auth.js
```diff
@@ -49,16 +50,22 @@ function getPathOrBase(site) {
  * @returns {Function}
  */
 function isAuthenticated(site) {
+  // console.log('~~~~~~~', site);
```
nelsonpecora: i hate this

lib/auth.test.js
```diff
@@ -82,6 +84,7 @@ describe(_.startCase(filename), function () {
           redirect: sandbox.spy()
         };
 
+      sites.getSiteFromPrefix.returns({ host: 'domain.com', prefix: 'domain.com' });
```
nelsonpecora: plz add a test

PR #434, opened 2017-08-31
serialize and deserialize user data in sessions

nelsonpecora:
* uses the key, fetches user data from the db every time
* this allows authenticating and deauthenticating without forcing the user to re-login
* this means smaller session records, which is good
* this requires all current sessions to be wiped, which is fine since we need to do that anyway to release amphora 4.x


nelsonpecora:
👍  from jon

PR #437, opened 2017-09-30
Amphora v5

jonwinton:
**Breaking Changes**
- All core routes now prefixed with an underscore (`_`)
    - `/_pages/`
    - `/_components/`
    - `/_uris/`
    - `/_users/`
    - `/_lists/`
- All keys in the DB are swapped to prefix with a site's slug rather than the site prefix
- Internal logging service swapped to use [`clay-log`](https://github.com/clay/clay-log)


nelsonpecora:
Kiln support for this change was added in version 4.6.0

pfeytser:
https://github.com/clay/amphora/issues/436

jonwinton:
Not doing db changes (yet)

PR #441, opened 2017-10-04
Clay Log Integration

jonwinton:
**Includes**
- Integrated `clay-log`
- Removed references to old logging module
- Old logging module is still exported, but there is a deprecation warning about Amphora 5


lib/render.js
```diff
@@ -340,11 +342,11 @@ function renderExpressRoute(req, res, next) {
       const diff = process.hrtime(hrStart),
         ms = Math.floor((diff[0] * 1e9 + diff[1]) / 1000000);
 
-      log('info', 'rendered express route:' + (req.hostname + req.baseUrl + req.path) + ' ' + ms + 'ms');
+      log('info', `rendered express route: ${req.hostname + req.baseUrl + req.path} ${ms}ms`);
     }
   }).catch((error) => {
     if (error.name === 'NotFoundError') {
-      log('verbose', 'in', req.uri, error.message);
+      log('error', `in ${req.uri}: ${error.message}`);
```
dmabuada: does error.message tell me it's not found? or is that only in error.name?

lib/bootstrap.js
```diff
@@ -265,10 +269,11 @@ function bootstrapComponents(site) {
 module.exports = function () {
   const sites = siteService.sites();
 
+  log('info', 'Beginning bootstrap');
```
dmabuada: does this need to be logged? if so, a success message would help too

lib/responses.js
```diff
@@ -289,7 +290,7 @@ function notFound(err, res) {
  */
 function serverError(err, res) {
   // error is required to be logged
-  log('error', err);
+  log('error', err.message);
```
dmabuada: consistent syntax plz

lib/render.js
jonwinton: It will tell you it's not found

lib/services/components.js
```diff
@@ -193,9 +195,9 @@ function put(uri, data, locals) {
       const ms = timer.getMillisecondsSince(startTime);
 
       if (ms > timeoutLimit * 0.5) {
-        log('warn', 'slow put', uri, ms + 'ms');
+        log('warn', `slow put ${uri} ${ms}ms`);
       }
-    }).timeout(timeoutLimit, 'Component module PUT exceeded ' + timeoutLimit + 'ms:' + uri);
+    }).timeout(timeoutLimit, `Component module PUT exceeded ${timeoutLimit}ms: ${uri}`);
```
dmabuada: can we log this timeout?

lib/bootstrap.js
jonwinton: Good call, one minute

lib/services/notifications.js
```diff
@@ -30,7 +32,7 @@ function callWebhook(event, url, data) {
   }
 
   return rest.fetch(url, options).then(function (response) {
-    log('info', 'called webhook', _.pick(response, ['status', 'statusText']));
+    log('info', `called webhook ${_.pick(response, ['status', 'statusText'])}`);
```
dmabuada: do we need to log allllll resp statuses?

lib/services/pages.js
```diff
@@ -291,11 +293,11 @@ function publish(uri, data, locals) {
...
 
       if (ms > timeoutLimit * 0.5) {
-        log('warn', 'slow publish', uri, ms + 'ms');
+        log('warn', `slow publish ${uri} ${ms}ms`);
       } else {
-        log('info', 'published', references.replaceVersion(uri), ms + 'ms');
+        log('info', `published ${references.replaceVersion(uri)} ${ms}ms`);
       }
-    }).timeout(timeoutLimit, 'Page publish exceeded ' + timeoutLimit + 'ms:' + uri);
+    }).timeout(timeoutLimit, `Page publish exceeded ${timeoutLimit}ms: ${uri}`);
```
dmabuada: here too.

lib/services/pages.test.js
```diff
@@ -10,25 +10,26 @@ const _ = require('lodash'),
   notifications = require('./notifications'),
   sinon = require('sinon'),
   siteService = require('./sites'),
-  timer = require('../timer'),
-  winston = require('winston');
+  timer = require('../timer');
```
dmabuada: does winston need to stay for backwards compatibility?

test/fixtures/api-accepts.js
```diff
@@ -341,6 +340,8 @@ function stubFiles(sandbox) {
   files.getComponentPath.withArgs('missing').returns('missingThing');
   files.getComponentPath.withArgs('invalid').returns(null);
 
+  sandbox.stub(files, 'getComponents').returns(['clay-c5', 'clay-c3', 'clay-c4']);
```
dmabuada: i have no idea what this does...but this file removes logging?

jonwinton: No, it just makes the function `getComponents` return this array when called from API tests.

lib/services/pages.test.js
jonwinton: Yes

test/fixtures/api-accepts.js
jonwinton: Oh, it removes the stub for the old logger from the test file, not from anything else. API tests are looking for the response objects, not what was logged. Individual files do that.

lib/services/notifications.js
jonwinton: It only logs `status` and `statusText`

dmabuada: i know, but would it be more worthwhile to log non-200 resps?

PR #448, opened 2017-10-06
Rename Bootstrap Property

jonwinton:
Renaming from `skipBootstrap` to `bootstrap`. Hasn't been released yet, so not a breaking change

### DIFF
--- a/lib/bootstrap.js
+++ b/lib/bootstrap.js
@@ -332,8 +332,8 @@ function skipBootstrap() {
   return bluebird.resolve();
 }
 
-module.exports = function (skip) {
-  return skip ? skipBootstrap() : bootrapSitesAndComponents();
+module.exports = function (skip = true) {
+  return skip ? bootrapSitesAndComponents() : skipBootstrap();
 };
 
 module.exports.bootstrapPath = bootstrapPath;
--- a/lib/bootstrap.js
+++ b/lib/bootstrap.js
@@ -332,8 +332,8 @@ function skipBootstrap() {
   return bluebird.resolve();
 }
 
-module.exports = function (skip = true) {
-  return skip ? bootrapSitesAndComponents() : skipBootstrap();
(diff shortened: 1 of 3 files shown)


lib/bootstrap.js
```diff
@@ -332,8 +332,8 @@ function skipBootstrap() {
   return bluebird.resolve();
 }
 
-module.exports = function (skip) {
-  return skip ? skipBootstrap() : bootrapSitesAndComponents();
+module.exports = function (skip = true) {
```
nelsonpecora: you should maybe rename this variable...

lib/bootstrap.test.js
```diff
@@ -75,15 +75,8 @@ describe(_.startCase(filename), function () {
...
-    it('skips boostrapping if the `skip` bool is true', function () {
-      return fn(true)
-        .then(function () {
-          sinon.assert.calledWith(fakeLog,'info', 'Skipping bootstrapping sites and components');
-        });
-    });
-
-    it('skips boostrapping if the `skip` bool is true', function () {
-      return fn(true)
+    it('skips bootstrapping if the `skip` bool is false', function () {
```
nelsonpecora: read this line, then think about it

lib/setup.js
```diff
@@ -30,7 +30,7 @@ module.exports = function (options = {}) {
       plugins = [],
       env = [],
       cacheControl = {},
-      skipBootstrap = false
+      bootstrap = true
```
nelsonpecora: will this be overwritten by a falsy value...?

jonwinton: Yes, destructuring default values are only used if a value is not present in the object. Here is a little easy reading on destructuring: http://wesbos.com/destructuring-default-values/

The `Careful: null and undefined` section outlines this behavior, but since `false`/`null`/`undefined`/`0` are still values, they would be taken instead of `true`

PR #453, opened 2017-10-12
Exclude components using query parameter

### LINKED ISSUE #452: Enable excluding components via query parameter
Enable users and machines to exclude components using query parameters.

For example, we could use the following search string to exclude all `block-quote`s and a
single instance of `clay-paragraph`:  `?_exclude=/block-quote/&_exclude=/clay-paragraph/instances/cj8nij4010000ys1u6yz6e48x`.

mheiber:
Fix #452 

Add ability to exclude components from
rendering using query parameters. This can be
helpful for testing page performance with different
combinations of components.

To use this feature:

- Add this key to your site config: `excludeUsingQueryParam`
and set its value to the name of the query parameter. For example:
`excludeUsingQueryParam: _exclude`.

- Restart Clay

- In a browser, request a page, using the new URL parameter.
For example, if the parameter is named `_exclude` and you want
to exclude the `block-quote` component, you can use a URL that ends with
`?_exclude=/block-quote/`. The matching is done using
String.prototype.includes, so you can match against the entire ref
for components or just part of the ref. You can also pass the parameter
multiple times. This search string excludes all `block-quote`s and a
single instance of `clay-paragraph`
`?_exclude=/block-quote/&_exclude=/clay-paragraph/instances/cj8nij4010000ys1u6yz6e48x`.

### DIFF
--- a/lib/services/composer.js
+++ b/lib/services/composer.js
@@ -7,6 +7,48 @@ const _ = require('lodash'),
   mapLayoutToPageData = require('../utils/layout-to-page-data'),
   referenceProperty = '_ref';
 
+
+/**
+ * Enables us to exclude certain components based on query parameter.
+ *
+ * Returns a function that can filter reference objects
+ * based on `locals`.
+ *
+ * `article` matches `/components/article/instances/foo` as well as `/components/layout/instances/article`,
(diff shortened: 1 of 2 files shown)


mheiber:
We're going to try to put this into an amphora-html plugin instead

PR #457, opened 2017-10-17
Pass missing auth level error to client

amycheng:
*Problem*: if there an auth level is not set for the user account, an error is thrown. This occurs if the user was bootstrapped incorrectly. The client will get a `500: Internal Server Error` error, which is not helpful, especially if the developer does not have access to Amphora logs

*Solution*:
- update `responses.serverError()` to send a custom [`response.statusMessage`](https://nodejs.org/dist/latest-v6.x/docs/api/http.html#http_response_statusmessage)
- catch the error so that `.serverError()` can be called

### DIFF
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -49,12 +49,18 @@ function checkAuthLevel(userLevel, requiredLevel) {
  */
 function withAuthLevel(requiredLevel) {
   return function (req, res, next) {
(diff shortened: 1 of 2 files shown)


amycheng:
@jonwinton suggested that make this solution work for more use cases, so that's what I'm going to work on.

jonwinton:
It doesn't seem like any further work is being done on this approach for now, closing to clear board. Can re-open later when a larger error handling architecture discussion arises.

PR #476, opened 2018-01-05
Stream Plugins

jonwinton:
- Expose a Stream API for plugins to use.
- Fixes mutation bug with competing publish/save events

### DIFF
--- a/lib/plugins/index.js
+++ b/lib/plugins/index.js
@@ -24,7 +24,8 @@ function executeHook(key, args) {
     // If we have a Stream for this plugin, write to it
     // but clone the ops to prevent bad mutations
     if (h.isStream(plugin[streamApi])) {
-      plugin[streamApi].write(_cloneDeep(args));
+      console.log(args);
+      h.of(_cloneDeep(args)).pipe(plugin[streamApi]);
     }
   });
 }
--- a/lib/plugins/index.js
+++ b/lib/plugins/index.js
@@ -24,8 +24,7 @@ function executeHook(key, args) {
     // If we have a Stream for this plugin, write to it
     // but clone the ops to prevent bad mutations
     if (h.isStream(plugin[streamApi])) {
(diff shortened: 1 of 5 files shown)


lib/plugins/index.js
```diff
@@ -1,6 +1,9 @@
 'use strict';
 
-const _ = require('lodash');
+const h = require('highland'),
+  _map = require('lodash/map'),
+  _isFunction = require('lodash/isFunction'),
+  _cloneDeep = require('lodash/cloneDeep');
```
nelsonpecora: Just a question, but any reason you're importing each separately rather than just `require('lodash')`? (since it's server-side)

lib/plugins/index.js
```diff
@@ -11,10 +14,19 @@ const _ = require('lodash');
  * @param  {Any} args
  */
 function executeHook(key, args) {
-  _.map(module.exports.plugins, (plugin) => {
-    if (_.isFunction(plugin[key])) {
+  _map(module.exports.plugins, (plugin) => {
```
cperryk: This should just be an `_.each` or `.forEach`

jonwinton: Oh that is true, will update. I wonder why I had it there before...Hmmm

cperryk: This should just be an `_.each`

nelsonpecora: Good point

lib/plugins/index.js
jonwinton: I'm starting to look the cost benefit of the how just importing `_` requires a running `Object.get` to run each time the function runs. I know V8 probably optimizes the calls so it doesn't really matter, I was just playing around.

ALSO it's on my list to remove lodash as much as possible (especially in the conversion of larger sections of the code to streams). It's nice to look at a file and get an initial glimmer at what it would take to refactor lodash out. I'll probably take this back to the original way, though.

PR #480, opened 2018-01-18
renderer models

jonwinton:
## Problem
The aim of this PR is to address rendering content for distributed networks (Apple News, RSS, AMP, etc). Clay is a system who is geared primarily towards serving/editing component data via a web browser, but this creates constraints on how data is saved/rendered when trying to use the same components for distributed networks.

## Solution
This problem was initially patched by making Amphora focused on JSON as the primary data delivery format and allowing renderers to handle the compilation of that data into whatever form is necessary. This initial approach worked well for serving HTML, but created one big issue: data is still composed in the same tree no matter what renderer you choose to use. Essentially, this means that if you need to format data deep in the tree, you have to write code in your renderer to traverse the tree, modify the data and either preserve or completely destroy the tree as Amphora understands it.

While we're not going to change how the tree is composed, what we can do is allow components to translate their own data into a format more friendly to the renderer as it leaves the DB. This means that component directories can now store a `<RENDERER_EXTENSION>.model.js` which will invoked _AFTER_ the default `model.js` file.

```
GET www.mysite.com/components/foo/instances/bar.rss

1. `foo/model.js` runs
2. Amphora checks for an `rss.model.js`
3. If found, Amphora `foo/rss.model.js` runs
4. Once data is composed, it's passed to the renderer for final compilation
```
From there a renderer can still do whatever it needs to do to handle the data. It can look for a template to handle formatting of the data, it can further transform the data and then respond with JSON. The flow is up to the author of the renderer.

## Caveat
The `html` extension is ignored. i.e. there are no `html.model.js` affordances. The reasoning goes back to Clay's primary outlet being the web browser. Your regular `model.js` should do everything it needs to do for formatting the data in the way Kiln will edit the component, these additional `<renderer>.model.js` files are for non-html renderers only.

## Renderer Model.js Files
These files should closely mirror the regular `model.js` files, except they only need to export one function by default.

```
module.exports = function (ref, data, locals) {

// Transform data
// Return a Promise which resolves data 
// or an object directly

return data;
};
```

## Changes

- Runs components through a renderer-specific `<RENDERER_EXTENSION>.model.js` file
- Updates `getComponentModule` to accept a `prefix` argument for model.js files
- Updates memoization for the `getComponentModule` file to support multiple arguments
- Adds the request extension to `locals`. When requesting straight data this value id `undefined`, only when requesting a page where a renderer would affect output
- Bugfix: Adds `locals` to the `get` of the layout data when rendering a page. This means the model.js file will actually run for a layout
- Remove bad `console.log` artifact

### DIFF
--- a/lib/files.test.js
+++ b/lib/files.test.js
@@ -292,6 +292,27 @@ describe(_.startCase(filename), function () {
 
       expect(fn(name)).to.equal(result);
     });
+
+    it('handles renderer model.js path', function () {
+      const name = 'some name',
+        path = 'some path',
+        prefix = 'foobar',
+        result = 'some result';
+
+      lib.getComponentPath.returns(path);
+      req.resolve.withArgs(`${path}/${prefix}.model`).returns(`${path}/${prefix}.model`);
+      req.withArgs(`${path}/${prefix}.model`).returns(result);
+
+      expect(fn(name, prefix)).to.equal(result);
+    });
+
+    it('returns undefined if the prefix is html', function () {
+      const name = 'some name',
+        prefix = 'html';
+
+      expect(fn(name, prefix)).to.be.false;
+      sinon.assert.notCalled(lib.getComponentPath);
+    });
   });
 
   describe('getComponentPackage', function () {
--- a/lib/files.test.js
+++ b/lib/files.test.js
@@ -306,7 +306,7 @@ describe(_.startCase(filename), function () {
       expect(fn(name, prefix)).to.equal(result);
     });
 
-    it('returns undefined if the prefix is html', function () {
+    it('returns false if the prefix is html', function () {
       const name = 'some name',
         prefix = 'html';
 
--- a/lib/services/components.js
+++ b/lib/services/components.js
@@ -85,7 +85,7 @@ function get(uri, locals) {
   }
 
   if (renderModel) {
-    promise.then(function (data) {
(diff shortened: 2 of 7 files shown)


lib/services/components.js
```diff
@@ -82,6 +84,12 @@ function get(uri, locals) {
     promise = db.get(uri).then(JSON.parse).then(upgrade.init(uri, locals)); // Run an upgrade!
   }
 
+  if (renderModel) {
+    promise.then(function (data) {
```
cperryk: Should be `promise = promise.then`

lib/files.test.js
```diff
@@ -292,6 +292,27 @@ describe(_.startCase(filename), function () {
...
+        result = 'some result';
+
+      lib.getComponentPath.returns(path);
+      req.resolve.withArgs(`${path}/${prefix}.model`).returns(`${path}/${prefix}.model`);
+      req.withArgs(`${path}/${prefix}.model`).returns(result);
+
+      expect(fn(name, prefix)).to.equal(result);
+    });
+
+    it('returns undefined if the prefix is html', function () {
```
cperryk: It returns `false`, not `undefined`

jonwinton: Oh yeah, missed that change

jonwinton:
Oh, and resolves https://github.com/clay/amphora/issues/429

PR #484, opened 2018-01-29
Unpub Error Logging

jonwinton:
Error notification for unpublished are not currently routed through the logging service.

### DIFF
--- a/lib/services/notifications.js
+++ b/lib/services/notifications.js
@@ -32,8 +32,18 @@ function callWebhook(event, url, data) {
   }
 
   return rest.fetch(url, options)
-    .then(function (response) {
-      log('info', `called webhook ${_.pick(response, ['status', 'statusText'])}`);
+    .then(function (resp) {
+      var { status, statusText } = resp;
+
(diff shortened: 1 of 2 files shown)


lib/services/notifications.js
```diff
@@ -31,9 +31,13 @@ function callWebhook(event, url, data) {
     options.body = data;
   }
 
-  return rest.fetch(url, options).then(function (response) {
-    log('info', `called webhook ${_.pick(response, ['status', 'statusText'])}`);
-  });
+  return rest.fetch(url, options)
+    .then(function (response) {
+      log('info', `called webhook ${_.pick(response, ['status', 'statusText'])}`);
```
cperryk: This doesn't relate to these changes specifically, but I have noticed that `node-fetch` doesn't terminate http agents (i.e. free up sockets) immediately unless responses are actually read (`res.buffer()`, `res.text()`, `res.json()`).

jonwinton: Oh, that's right. Let me add a thing for that.

PR #505, opened 2018-02-02
V5

jonwinton:
# API Changes
- Underscores routes
- Renderers now terminate the Express response
- Removed references to `resolveMedia`
- Removed the check for memory leaks from `control`
- Deprecate the old publish API

# Other Changes
- Circle tests on 8.9.4
- Removed the `winston` logger completely from tests


lib/render.test.js
```diff
@@ -183,18 +148,18 @@ describe(_.startCase(filename), function () {
...
-      db.get.returns(resolveString({layout: '/components/hey'}));
+      db.get.returns(resolveString({layout: '/_components/hey'}));
 
       // 2. get the data that will be composed in the template
       components.get.returns(bluebird.resolve({}));
 
-      // 3. get the template that will be composed
-      components.getTemplate.returns('asdf');
+      // // 3. get the template that will be composed
+      // components.getTemplate.returns('asdf');
```
byronhulcher: should this (and other commented out lines from this PR) be removed from the test file? Hard to tell if a commented out part of a test is temporarily broken or purposeful removal

jonwinton: Oh, good find. Forgot that was there.

lib/services/publish.test.js
```diff
@@ -7,11 +7,10 @@ const _ = require('lodash'),
   db = require('../services/db'),
   lib = require(`./${filename}`);
 
-require('sinon-as-promised');
+// require('sinon-as-promised');
```
byronhulcher: can be removed?

jonwinton: Yep! Will also remove. So much deprecation. This is a reminder to myself to also check that I removed `winston`

PR #506, opened 2018-02-05
Renderers Terminate Response

jonwinton:
- Removes `resolveMedia` from the sites controller
- Kills the media service
- Removes some old functions around finding templates that should have been removed awhile ago


lib/render.js
```diff
@@ -201,38 +191,11 @@ function renderComponent(req, res, options) {
...
-        renderTime: ms,
-        route,
-        type: resp.type
-      });
-
-    }
-
-    return resp;
-  };
+        .then(state => renderer.render(state, res));
```
cperryk: We can move this to the outside level -- no need for a second promise chain

lib/services/components.js
```diff
@@ -523,12 +435,7 @@ module.exports.del = del;
 module.exports.post = post;
 
 // repeatable look-ups
-module.exports.getName = getComponentName; // too large to memoize
-module.exports.getTemplate = getTemplate; // too large to memoize
-module.exports.getPossibleTemplates = control.memoize(getPossibleTemplates, joinArguments);
 module.exports.getSchema = control.memoize(getSchema);
-module.exports.getScripts = control.memoize(getScripts, joinArguments);
-module.exports.getStyles = control.memoize(getStyles, joinArguments);
```
cperryk: 🎆

PR #507, opened 2018-02-05
Deprecate publish

jonwinton:
Deprecates old publish API


lib/render.js
```diff
@@ -344,6 +344,7 @@ function renderExpressRoute(req, res, next) {
 
   return module.exports.renderUri(pageReference, req, res)
     .catch((error) => {
+      // console.log(error.stack);
```
cperryk: Bloop

lib/services/publish.js
```diff
@@ -169,26 +160,18 @@ function resolvePublishUrl(uri, locals, site) {
...
 
     if ( publishingChain.length > 0 ) {
       // Iterate over an array of publishing functions sequentially to find the first one which resolves
       return chain(publishingChain, uri, _.cloneDeep(pageData), locals)
-        .then(function (urlOrPageData) {
-          return useResolvePublishUrl ? publishPageAtUrl(urlOrPageData, uri, pageData, locals, site) : urlOrPageData;
-        })
+        .then(url => publishPageAtUrl(url, uri, pageData, locals, site))
         .catch(e => {
           throw e;
```
cperryk: While we're in here, this catch handler doesn't do anything

PR #519, opened 2018-02-27
get component variations by reading css files

amycheng:
... and then set the state with them

related to https://github.com/clay/amphora-html/issues/25


amycheng:
moving to `amphora-html`

PR #522, opened 2018-03-12
Expose stack in server errors responses

cperryk:


lib/responses.js
```diff
@@ -292,7 +292,7 @@ function serverError(err, res) {
   // error is required to be logged
   log('error', err.message, { stack: err.stack });
 
-  const message = err.message || 'Server Error', // completely hide these messages from outside
+  const message = err.stack, // completely hide these messages from outside
```
nelsonpecora: that comment tho...

nelsonpecora:
I think the tests need to be updated

nelsonpecora:
I think the tests need to be updated tho...

PR #524, opened 2018-03-16
Breaking Change: Consolidating Render Data

jonwinton:
Regarding https://github.com/clay/amphora/issues/521

Building a separate rendering node is currently impossible simply because of how much work Amphora is doing to compile data for Kiln to pass to Amphora HTML. To begin down the road for separating the two, we need to move even more rendering specific logic out of Amphora.

Requirements:

- Amphora FS module 
- Support changes in Amphora renderers (with amphora-fs) module

# Breaking Changes
- Moved `getComponents`, `getComponentName`, `getComponentPath`, `getIndices` and `getSchemaPath` into amphora-fs and out of Amphora
- Renderer function signature changed to three arguments:
    - The composed data for the page/component
    - An object with `locals`, the ref being rendered and the ref of the layout (if it's a page)
    - The Express response object 
- Amphora no longer accepts `envVar` JSON args. This was used only for rendering HTML and will be supported by Amphora HTML and any other renderer that needs it


lib/files.test.js
```diff
@@ -38,16 +37,14 @@ describe(_.startCase(filename), function () {
...
-    lib.getComponentPath.cache = new _.memoize.Cache();
     lib.getComponentModule.cache = new _.memoize.Cache();
     lib.getComponentPackage.cache = new _.memoize.Cache();
     lib.isDirectory.cache = new _.memoize.Cache();
     lib.fileExists.cache = new _.memoize.Cache();
     lib.tryRequire.cache = new _.memoize.Cache();
 
     // package file can never be real
-    lib.setPackageConfiguration(pkg);
+    // lib.setPackageConfiguration(pkg);
```
cperryk: ☝️

Issue #500, opened 2018-02-01
"plugins"


nelsonpecora:
* plugins
* renderers
* differences between them
* links to api docs

PR #536, opened 2018-05-03
Plugin Docs

jonwinton:
#500

### DIFF
--- a/docs/plugins/hooks.md
+++ b/docs/plugins/hooks.md
@@ -1,0 +1,98 @@
+# Hooks
+
+Below are details about the plugin hooks in Amphora. Plugin hooks are fired on each plugin supplied to Amphora at instantiation time. A plugin should expose a property whose name corresponds to one of the hooks below and is a function that expects the arguments detailed below.
+
+---
+
+### Routes (`routes`)
+
+A hook that gets called with the argument of the [Express Router](https://expressjs.com/en/4x/api.html#express.router) for each site in a Clay instance. This hooks allows a plugin to attach a route to a site that can then be called from a client.
+
+**Arguments:**
+  - `router`. An instance of the [Express Router](https://expressjs.com/en/4x/api.html#express.router) for a site
+
+---
+
+### Save (`save`)
+
+A hook that is called with an Array of DB operations that were just processed from Clay data structure(s) being saved.
+
+**Arguments:**
+  - `ops`: an array of DB operations
+
+---
+
+### Delete (`delete`)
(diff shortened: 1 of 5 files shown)


docs/plugins/plugin-vs-renderer.md
```diff
@@ -0,0 +1,15 @@
+# Plugin vs. Renderer
```
byronhulcher: Thank you for this 👍

docs/plugins/hooks.md
```diff
@@ -0,0 +1,98 @@
+# Hooks
+
+Below are details about the plugin hooks in Amphora. Plugin hooks are fired on each plugin supplied to Amphora at instantiation time. A plugin should expose a property whose name corresponds to one of the hooks below and is a function that expects the arguments detailed below.
+
+---
+
+### Routes (`routes`)
+
+A hook that gets called with the argument of the [Express Router](https://expressjs.com/en/4x/api.html#express.router) for each site in a Clay instance. This hooks allows a plugin to attach a route to a site that can then be called from a client.
```
cperryk: `this hooks allows`

docs/plugins/hooks.md
```diff
@@ -0,0 +1,98 @@
...
+  - `router`. An instance of the [Express Router](https://expressjs.com/en/4x/api.html#express.router) for a site
+
+---
+
+### Save (`save`)
+
+A hook that is called with an Array of DB operations that were just processed from Clay data structure(s) being saved.
+
+**Arguments:**
+  - `ops`: an array of DB operations
```
cperryk: Can we give an example somewhere of what one of these operations look like?

docs/plugins/writing-a-plugin.md
```diff
@@ -0,0 +1,33 @@
...
+// code before amphora instantiation...
+amphora({
+  plugins: [
+    myPlugin
+  ]
+}).then(...)
+```
+---
+
+This example leaves out the proper Amphora instantiation in favor of showing how to write a very simple plugin and pass it to Amphora. The key takeaway is that plugins are objects whose top level properties correspond to the names of [Amphora hooks](hooks.md) with the value of those properties being function handlers.
```
byronhulcher: if `plugins` is an array, can I attach multiple hooks with the same name? Or will they over-write the previously registered plugin?  If you could clarify that it would be helpful.

docs/plugins/hooks.md
jonwinton: That should be handled in a different doc about how to hook up whichever key/val store you want.

PR #545, opened 2018-05-23
call publishLayout hook when publishing layouts

nelsonpecora:
* this implements the amphora portion of https://github.com/clay/clay-kiln/issues/1105
* amphora will look at component schemas (saved in memory) to see if they have a root-level `_layout` boolean. If so, it will execute the `publishLayout` hook
* warn when saving pages that reference layouts that _don't_ have the `_layout` property (we'll make this an error in the next major version)

### DIFF
--- a/lib/plugins/readme.md
+++ b/lib/plugins/readme.md
@@ -80,3 +80,7 @@ Note that `publish` and `unpublish` are used by certain plugins, but they will b
   - hook triggered on page unscheduling
   - _note:_ when `unschedulePage` is triggered, `save` is also triggered
   - given one param with uri, data, and user: `{ uri: scheduledItemUri, data: { at: timestamp, publish: pageUri }, user: req.user }`
+
+### Layout-specific Methods
+
+Layouts only have one hook, called `publishLayout`. It is called with the same arguments as `publishPage`, namely `{ uri, data, user }`.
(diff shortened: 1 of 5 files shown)


lib/plugins/readme.md
```diff
@@ -80,3 +80,7 @@ Note that `publish` and `unpublish` are used by certain plugins, but they will b
   - hook triggered on page unscheduling
   - _note:_ when `unschedulePage` is triggered, `save` is also triggered
   - given one param with uri, data, and user: `{ uri: scheduledItemUri, data: { at: timestamp, publish: pageUri }, user: req.user }`
+
+### Layout-specific Methods
+
+Layouts only have one hook, called `publishLayout`. It is called with the same arguments as `publishPage`, namely `{ uri, data, user }`.
```
byronhulcher: I'd keep out nix "It is called with the same arguments as `publishPage`" in case you forget to remove that if the arguments for publishPage change and this doesn't.  Just including the args should suffice!

nelsonpecora: I really don't want those to diverge though! This comment establishes that intent

nelsonpecora:
Ehhh, `layout` is a perfectly legit field name for a component to have 🙃

nelsonpecora:
The cool thing is that the `layouts` index is super fault-tolerant, so if you save / publish a layout that doesn't currently exist (because you forgot to add `_layout` before, or whatever) it'll create the elastic doc and update it correctly.

PR #546, opened 2018-06-04
publishing dynamic page

jonwinton:
- Publishes a dynamic page if a `_dynamic` property is set to `true` on the page data
- If dynamic, a uri is not created
- Attaches `publishUrlDynamic` to `locals` to mimic `publishUrl`. It's very handy for processing data in a model.js `render` for a component when publishing

### DIFF
--- a/lib/services/publish.js
+++ b/lib/services/publish.js
@@ -109,6 +109,22 @@ function getPassthroughPageUrl(uri, { url, customUrl }) {
   return bluebird.resolve(customUrl || url);
 }
 
+/**
+ * Allow a page to be published if it is dynamic. This allows the page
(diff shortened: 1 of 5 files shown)


lib/services/publish.js
```diff
@@ -125,6 +141,11 @@ function publishPageAtUrl(url, uri, data, locals, site) { // eslint-disable-line
   // implementation, but customUrl should override those rules
   url = data.customUrl || url;
 
+  if (data._dynamic) {
+    locals.publishUrlDynamic = true;
```
nelsonpecora: publishUrl is the actual url, right? this seems confusing. maybe `isDynamicPublishUrl` if it's gonna be a bookean?

PR #547, opened 2018-06-04
render from published  instance

jonwinton:
Fixes bug with dynamic pages rendering from the unpublished instance

### DIFF
--- a/lib/render.js
+++ b/lib/render.js
@@ -133,7 +133,7 @@ function renderDynamicRoute(pageId) {
   return (req, res) => {
     const hrStart = process.hrtime(),
       site = res.locals.site,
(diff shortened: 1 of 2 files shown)


lib/render.js
```diff
@@ -133,7 +133,7 @@ function renderDynamicRoute(pageId) {
   return (req, res) => {
     const hrStart = process.hrtime(),
       site = res.locals.site,
-      pageUri = `${getExpressRoutePrefix(site)}/_pages/${pageId}`;
+      pageUri = `${getExpressRoutePrefix(site)}/_pages/${pageId}@published`;
```
nelsonpecora: ah, useful

PR #554, opened 2018-06-21
New DB API: Storage

jonwinton:
- Pulls out the DB module to use supplied `storage`
- Refactors reads from DB to ALWAYS expect JSON, not strings of JSON

Issue: https://github.com/clay/amphora/issues/553


lib/responses.js
```diff
@@ -503,13 +503,15 @@ function listWithPublishedVersions(req, res) {
...
+  expectJSON(() => {
+    var promise = db.pipeToPromise(db.list({
       keys: true,
       values: true,
       isArray: true,
       prefix: req.uri
-    })).then(JSON.parse).then(convertDBListToReferenceObjects);
+    }));
+
+    return promise.then(JSON.parse).then(convertDBListToReferenceObjects);
```
nelsonpecora: does `pipeToPromise` still return strings of JSON?

lib/services/db.js
```diff
@@ -148,146 +107,15 @@ function list(options) {
...
-module.exports.batch = batch;
-module.exports.clear = clear;
-module.exports.getDB = function () { return db; };
-module.exports.setDB = function (value) { db = value; };
 module.exports.pipeToPromise = pipeToPromise;
+module.exports.registerStorage = registerStorage;
 
-Eventify.enable(module.exports);
-
-_.each(['put', 'batch', 'del'], exposeMethodToOutsideListeners);
```
nelsonpecora: what outside listeners did we have (if any)?

jonwinton: None, we just never deprecated it 😬

lib/responses.js
jonwinton: Unfortunately yes because the stream requires strings to operate, so we have to parse it out. I've done a little refactor on this and will push it up soon. The only thing using method is scheduling, so when that gets pulled out this is just going to move over

PR #582, opened 2018-08-07
Amphora FS

jonwinton:
Moves the files service to use [Amphora FS](https://www.npmjs.com/package/amphora-fs). We need to memoize a couple though, so handle that.


lib/files.test.js
```diff
@@ -1,323 +1,323 @@
...
-      });
-    });
-  });
-});
+//       fn('.').catch(function (result) {
+//         expect(result).to.equal('');
+//       });
+//     });
+//   });
+// });
```
rmfarrell: can commented out code be 🔪 'ed? If you need to restore it it will always be in history

PR #598, opened 2018-08-22
Validate site routes

zonika:
1. Attaches Amphora's reserved api routes _first_ so requests won't get caught by a site's less specific route path definition (therefore throwing an error that isn't actually an error) if they're actually meant to get caught by Amphora
2. Validates site-specific route paths to prevent sites from overwriting Amphora api routes now that they're added after the api routes


lib/services/attachRoutes.js
```diff
@@ -1,6 +1,28 @@
 'use strict';
 
-const render = require('../render');
+const render = require('../render'),
+  _ = require('lodash');
+
+let reservedRoutes,
+  log = require('./logger').setup({ file: __filename });
+
+function validPath(path, site) {
```
jonwinton: Function description

lib/services/attachRoutes.js
```diff
@@ -1,6 +1,28 @@
...
+  let reservedRoute;
+
+  if (path[0] !== '/') {
+    log('warn', `Path '${path}' must begin with a slash, skipping route.`, { path, site: site.slug });
+    return false;
+  } else {
```
jonwinton: You shouldn't need this `else`

PR #607, opened 2018-09-12
Redirect From Published Meta

jonwinton:
Requests to `/_pages/:id@published/meta` are redirected to `/_pages/:id/meta` so that the same data is not accessible in two locations.

- Fixes lint issue
- Adds testing helper for testing redirects
- Redirect logic for layouts and pages

### DIFF
--- a/lib/responses.js
+++ b/lib/responses.js
@@ -524,11 +524,27 @@ function deleteRouteFromDB(req, res) {
   }, res);
 }
 
+/**
+ * Forces a redirect back to the non-published
+ * version of a uri
+ *
+ * @param {Number} code
+ * @returns {Function}
+ */
+function forceDefault(code = 303) {
+  return (req, res) => {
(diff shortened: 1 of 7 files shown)


lib/responses.js
```diff
@@ -524,12 +524,28 @@ function deleteRouteFromDB(req, res) {
...
 }
 
+/**
+ * Forces a redirect back to the non-published
+ * version of a uri
+ *
+ * @param {Number} code
+ * @returns {Function}
+ */
+function forceDefault(code = 303) {
```
rmfarrell: I feel this function name is confusing. "Canonical" seems more correct to me than "default"

jonwinton: Hmm I disagree as it's simply a redirect back the the editable/default page. I could go for `editable`, but this is not what I would call a canonical route. 

I'll change to `editable`

PR #609, opened 2018-09-28
Event Bus Plugin

jonwinton:
Allows a custom event bus to be passed in and invoked in the same way the built-in event bus is triggered.

In order to provide a more robust plugin API, the stringification of the bus message is moved to the bus service itself. It should be the choice of the plugin how to stringify the message.

### DIFF
--- a/lib/services/models.js
+++ b/lib/services/models.js
@@ -66,10 +66,6 @@ function get(model, renderModel, executeRender, uri, locals) { /* eslint max-par
 
     promise = bluebird.try(() => {
       return db.get(uri)
-        // .then(data => {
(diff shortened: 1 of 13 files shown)


lib/services/models.js
```diff
@@ -66,6 +66,10 @@ function get(model, renderModel, executeRender, uri, locals) { /* eslint max-par
 
     promise = bluebird.try(() => {
       return db.get(uri)
+        // .then(data => {
```
yuliyv: Probably should remove

jonwinton: Whoops that got in there.

PR #618, opened 2018-11-06
Return 404 error when targeting a non-existent instance of a component

labreu05:
**Description:**
- Return `404` error when targeting a non-existent instance of a component with a renderer.

**Context:**
- When a non-existent instance of a component is requested in the browser passing a renderer the request hangs and responds with an error `503`.

**Issue reference:**
- https://github.com/clay/amphora/issues/617

### DIFF
--- a/lib/render.js
+++ b/lib/render.js
@@ -87,13 +87,11 @@ function renderComponent(req, res, hrStart) {
   return components.get(_ref, locals)
     .then(cmptData => formDataForRenderer(cmptData, { _ref }, locals))
     .tap(logTime(hrStart, _ref))
-    .then(({data, options}) => renderer.render(data, options, res))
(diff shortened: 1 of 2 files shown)


lib/render.js
```diff
@@ -88,7 +88,13 @@ function renderComponent(req, res, hrStart) {
     .then(cmptData => formDataForRenderer(cmptData, { _ref }, locals))
     .tap(logTime(hrStart, _ref))
     .then(({data, options}) => renderer.render(data, options, res))
-    .catch(err => log('error', err));
+    .catch(err => {
+      const errorHandler = responses.handleError(res);
+
+      log('error', err);
+
+      return errorHandler(err);
```
jonwinton:
```suggestion
      return responses.handleError(res)(err);
```
And then you can remove the `const` declaration.

PR #626, opened 2019-01-30
Google+ Deprecation Support

jonwinton:
Solution pulled from: https://github.com/jaredhanson/passport-google-oauth2/pull/51

To use the new authentication API set `GOOGLE_PROFILE_URL` to `https://www.googleapis.com/oauth2/v3/userinfo`


jonwinton:
It should work with current keys, just tested locally.

PR #631, opened 2019-02-19
Amphora passport

manuelurenah:
## Description

Adds support for new amphora-passport module


lib/setup.js
```diff
@@ -30,7 +30,7 @@ module.exports = function (options = {}) {
   let {
       app = express(),
       providers = [],
-      sessionStore,
+      sessionStore, // TODO: add deprecation notice
```
jonwinton: @manuelurenah right below this declaration can you add a deprecation log (level `warn` that we'll be dropping support for the `sessionStore` property in Amphora v8?

package.json
```diff
@@ -17,6 +17,7 @@
   "license": "MIT",
   "dependencies": {
     "amphora-fs": "^1.0.1",
+    "amphora-passport": "file:../amphora-passport",
```
jonwinton: We'll need to do a release of the module for testing. Whenever that work is done can you release a `v0.0.1` of the package and we'll test it out?

manuelurenah:
Closing in favor for `amphora-authentication`

PR #635, opened 2019-03-06
Amphora authentication

manuelurenah:
### Description

* Moves all user handling logic from `amphora` into `amphora-authentication` project
* Removes test cases concerning user authentication
* Refactors routes file
* Code clean up

### QA Notes

* Do `npm install` to get `amphora-authentication` package
* Run `npm run test` to verify that all tests are passing and code coverage is at 100%
* Link the `amphora` module to a clay instance
* Add the `local` provider to the array of supported providers when setting up `amphora`
* Import a new user into your clay instance that supports local auth:
  ```yml
  -
    username: murena
    password: secret
    provider: local
    auth: admin
  ```
* Check that the user is able to login with username and password as well as with any other provider
* Verify that the `/_users` routes are working as expected (they shouldn't show the user password)

### Next Steps

* Update `amphora-authentication` module when a official release is out


lib/routes.js
```diff
@@ -148,10 +132,26 @@ function addControllerRoutes(router) {
...
+ * Add routes from the authentication module that
+ * can be passed in. If no module is passed in at
+ * instantiation time then use the default auth module.
+ *
+ * @param {express.Router} router
+ */
+function addAuthenticationRoutes(router) {
+  const authRouter = express.Router();
+
+  authRouter.use(require('body-parser').json({ strict: true, type: 'application/json', limit: '50mb' }));
```
james-owen: I've seen a couple of different numbers for the body size limit. I'm not sure how to figure out what it should be, but we should check this number.

lib/routes.js
```diff
@@ -218,17 +218,23 @@ function addSiteController(router, site, providers) {
...
+function addSite(router, { providers }, site) {
   site = _.defaults(site, { path: '/' }); // If path isn't explicitly set, set it.
   const path = site.path,
     siteRouter = express.Router();
 
   // authentication added first to protect all routes
   if (!_.isEmpty(providers)) {
-    auth.init(siteRouter, providers, site, sessionStore);
+    // Initialize auth module
+    amphoraAuth({
```
james-owen: Nice.

lib/routes.js.orig
```diff
@@ -0,0 +1,397 @@
+'use strict';
```
jonwinton: Good to delete this file?

lib/services/publish.test.js.orig
```diff
@@ -0,0 +1,296 @@
+'use strict';
```
jonwinton: And this one?

lib/setup.js.orig
```diff
@@ -0,0 +1,77 @@
+'use strict';
```
jonwinton: And this one?

lib/routes.js
```diff
@@ -218,17 +218,23 @@ function addSiteController(router, site, providers) {
...
+      router: siteRouter,
+      providers,
+      site,
+      storage: db,
+      bus
```
jonwinton: Is the auth module using the bus? I thought it was using the session store

jonwinton: Are we saying the session store is now instantiated in the auth module and is no longer required to be passed in? I think that would be a be a breaking change in that we would be taking over the ability of the user to customize their session store. What do you think about passing in a reference to the session store here and then in amphora-auth you can instantiate it if a session store if one wasn't passed in.

lib/routes.js
```diff
@@ -148,10 +132,26 @@ function addControllerRoutes(router) {
...
+ *
+ * @param {express.Router} router
+ */
+function addAuthenticationRoutes(router) {
+  const authRouter = express.Router();
+
+  authRouter.use(require('body-parser').json({ strict: true, type: 'application/json', limit: '50mb' }));
+  authRouter.use(require('body-parser').text({ type: 'text/*' }));
+  amphoraAuth.addRoutes(authRouter);
+  router.use('/_users', authRouter);
```
jonwinton: Nice scoping 👍

test/mocha.opts
```diff
@@ -1,4 +1,5 @@
 --reporter dot
 --ui bdd
 --slow 50
---timeout 200
\ No newline at end of file
+--timeout 200
+--exit
```
jonwinton: Reminder that we talked about leaving a comment to specify why this is necessary

lib/routes.js
manuelurenah: Yeah, I'm not sure either and I feel `50mb` is a lot, but I think the default value (`100kb`) should be enough or at most `1mb` for the body size limit. What do you think?

lib/routes.js
manuelurenah: Yes, we are using the bus to publish the save and delete action of users, please see [here](https://github.com/clay/amphora-auth/pull/11/files#diff-71dfd080cffe4efa09fc54e1b1b92ef4R33) and [here](https://github.com/clay/amphora-auth/pull/11/files#diff-71dfd080cffe4efa09fc54e1b1b92ef4R48)

lib/routes.js
james-owen: Here, we have 1mb for nginx and 5mb for Express: https://github.com/clay/clay-starter/blob/master/app/services/startup/index.js but let's check with @jonwinton on this.

lib/routes.js
manuelurenah: Also, since we are dropping support for session store in Amphora v8 ([see here](https://github.com/clay/amphora/pull/635/files#diff-f5d8c0f33aa88aa6e119a0cdd5f1c40eR44) for warning message) I thought we were going to handle this through amphora-auth. But that's a better idea to avoid breaking changes

jonwinton: Ah, I totally forgot about the bus for that. Good call. But yeah, we need to make sure not to drop support in v7. Nice job!

Issue #634, opened 2019-03-01
Amphora Redirects Do Not Preserve Query Params


jonwinton:
## Description

[When Amphora redirects a client](https://github.com/clay/amphora/blob/6.x-lts/lib/render.js#L219-L227) to a new url the query params are not preserved

## Steps to Reproduce

1. Publish a page
2. Change the url and re-publish
3. Go to the first url with a query param defined (that's not `?edit=true`)
4. See url in navigation bar

## Expected Behavior

The url contains the same query param as the initial request

## Additional Comments

It might be easier to just pull everything after the `?` from the request object rather than rebuilding from `req.query`, but we'll need to see how that interacts with the edit mode query param

PR #642, opened 2019-03-20
Preserve query parameters on redirect

james-owen:
#634


lib/render.test.js
```diff
@@ -227,12 +227,36 @@ describe(_.startCase(filename), function () {
...
+    it('maintains querystring on redirect', function () {
+      const resultRef = 'domain.com/_uris/zxccv',
+        mockRes = getMockRes(),
+        requestObj = {
+          _parsedUrl: {
+            search: '?query=string'
+          }
+        };
+
+      // console.log("\n\n\n\n MOCK REQ\n\n\n\n", mockReq);
```
jonwinton: Extra console.log here

lib/render.test.js
```diff
@@ -227,12 +227,36 @@ describe(_.startCase(filename), function () {
...
+
+      // when we find the new reference
+      db.get.returns(bluebird.resolve('domain.com/_uris/' + new Buffer(resultRef).toString('base64')));
+
+      return fn('/_uris/asdf', requestObj, mockRes).then(function () {
         expect(mockRes.redirect.callCount).to.equal(1);
```
jonwinton: Even though the test passes we're not actually testing if the `res.redirect` method was called with a query string attached to the url. 

I would suggest testing `mockRes.redirect` using `calledWith` to make sure the query param is present

Issue #613, opened 2018-10-17
Prevent Archived Publish


jonwinton:
## Description

When a page is archived it should not be able to be published via an API call.

## Steps to Reproduce

1. Create a new page
2. Archive a page
![Screen Shot 2019-03-21 at 8 54 15 AM](https://user-images.githubusercontent.com/4906712/54753409-5d6c7c80-4bb7-11e9-951b-9243931e39f0.png)
3. Execute a PUT against Amphora to publish the page
4. The page will publish

## Expected Behavior

The publish doesn't happen

## Additional Comments

Amphora accepts an array of functions that can determine how to publish any given page. The logic for processing these functions and the page data [can be found here](https://github.com/clay/amphora/blob/master/lib/services/publish.js). 

But amphora also handles two functions itself for [dealing with custom urls and dynamic pages](https://github.com/clay/amphora/blob/master/lib/services/publish.js#L218).

We should add a third rule that immediately checks the page for the `archived` property so that we can can cancel the publish quickly. The `archived` property is found on the page metadata.

jonwinton:
We should also not allow published articles to be programmatically archived without being unpublished first.

PR #644, opened 2019-03-26
Check archived property on publish

james-owen:
_checkForArchived function in publish.js

### DIFF
--- a/lib/services/publish.js
+++ b/lib/services/publish.js
@@ -107,19 +107,20 @@ function _checkForDynamicPage(uri, { _dynamic }) {
 }
 
 /**
+ * Checks the archive property on a page's meta object, to determine whether
+ * it can be published. Archived pages should not be published. Otherwise, an
+ * initial object is passed to the processing chain in processPublishRules.
+ *
  * @param {String} uri
- * @returns {Promise|Undefined}
+ * @returns {Promise}
  */
 function _checkForArchived(uri) {
-  // get meta with uri
   return db.getMeta(replaceVersion(uri))
     .then(meta => meta.archived)
-    .then(archivedBool => {
-      if (archivedBool) {
-        return bluebird.reject({val: '', errors: {_checkForArchived: 'The page is archived and cannot be published'}});
-      }
-      return bluebird.resolve({ val: '', errors: {}});
-    });
+    .then(archived => archived
+      ? bluebird.reject({ val: '', errors: { _checkForArchived: 'The page is archived and cannot be published' } })
+      : bluebird.resolve({ val: '', errors: {} })
+    );
 }
 
 /**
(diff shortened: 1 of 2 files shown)


james-owen:
#613

lib/services/publish.test.js
```diff
@@ -56,12 +56,32 @@ describe(_.startCase(filename), function () {
       rule1.returns(Promise.reject(new Error(0)));
       rule2.returns(Promise.resolve('foo.com/bar/baz'));
 
-      return fn([ () => Promise.reject(new Error(0)), rule2, rule3 ])
+      const uri = 'http://example.com/uri@published';
+
+      db.getMeta.returns(Promise.resolve({archived: false}));
```
manuelurenah: Please mind spacing
```suggestion
      db.getMeta.returns(Promise.resolve({ archived: false }));
```

lib/services/publish.test.js
```diff
@@ -92,6 +112,8 @@ describe(_.startCase(filename), function () {
       modifyFn.returnsArg(0);
       db.get.returns(Promise.resolve(JSON.stringify(fakePage)));
       meta.getMeta.returns(Promise.resolve({}));
+      db.getMeta.returns(Promise.resolve({archived: false}));
```
manuelurenah: Same suggestion as before

lib/services/publish.test.js
```diff
@@ -237,4 +265,32 @@ describe(_.startCase(filename), function () {
...
+
+    it('resolves with an empty error object if the meta.archived property is false', function () {
+      const uri = 'http://example.com/uri@published';
+
+      db.getMeta.returns(Promise.resolve({archived: false}));
+
+      return fn(uri)
+        .then(resp => {
+          expect(resp).to.eql({val:'', errors: {}});
+        });
```
manuelurenah: We could have this in one line
```js
.then(resp => expect(resp).to.eql({ val: '', errors: {} }));
```

lib/services/publish.test.js
```diff
@@ -237,4 +265,32 @@ describe(_.startCase(filename), function () {
...
+    const fn = lib[this.title];
+
+    it('rejects with an error object if the meta.archived property is true', function () {
+      const uri = 'http://example.com/uri@published';
+
+      db.getMeta.returns(Promise.resolve({archived: true}));
+
+      return fn(uri)
+        .catch(err => {
+          const expectedErr = {val:'', errors: {_checkForArchived: 'The page is archived and cannot be published'}};
```
manuelurenah: Please mind spacing when using objects. I know this is being picky 😛 but it helps for code style and readability
```suggestion
          const expectedErr = { val:'', errors: { _checkForArchived: 'The page is archived and cannot be published' } };
```

lib/services/publish.js
```diff
@@ -106,6 +106,22 @@ function _checkForDynamicPage(uri, { _dynamic }) {
   return bluebird.resolve(_dynamic);
 }
 
+/**
+ * @param {String} uri
+ * @returns {Promise|Undefined}
```
PedroAlbR: This method doesn't return `undefined`, we could remove it.

PR #650, opened 2019-04-09
Plugin whitelist

### LINKED ISSUE #640: Amphora Plugins Auth Requirement
## Description

Currently any Amphora plugin can only accept `GET` requests because of this line: https://github.com/clay/amphora/blob/master/lib/auth.js#L69

Reported by @jjpaulino

## Steps to Reproduce

1. Make any plugin which accepts a request other than `GET` or `OPTIONS`
2. Execute a request to the plugin
3. See requirement for authentication

## Expected Behavior

A plugin should be able to opt-out of authentication requirements

## Additional Context

We should make the authentication requirement a default but allow plugins to opt-out. This will require moving the function which is enforcing this to be applied individually to core routes and then a test for plugin routes will need to be added.

PedroAlbR:
# Whitelist to skip plugin auth

Plugins can also be objects now so they can pass metadata to amphora like a routes whitelist.

```js
...
plugins: [
  ...,
  amphoraSearchPlugin,
  { init: testPluginFunction, whitelist: ['/_test-route'] };
],
...
```
This allows amphora to know which routes should skip the authentication

## TODO
- [ ] Update [plugins documentation](https://github.com/clay/amphora/blob/9443c88ae5659c0d1693172db9d80d8b0eace477/docs/basics/plugins.md)
- [ ] Memoization of the route check

Closes #640

### DIFF
--- a/lib/services/plugins.test.js
+++ b/lib/services/plugins.test.js
@@ -11,12 +11,13 @@ const _ = require('lodash'),
   expect = require('chai').expect;
 
 describe(_.startCase(filename), function () {
-  let sandbox, plugin, fakeLog;
+  let sandbox, plugin, pluginWithMetadata, fakeLog;
 
   beforeEach(function () {
     sandbox = sinon.sandbox.create();
 
     plugin = sandbox.stub();
+    pluginWithMetadata = { init: sandbox.stub().returns(true) };
     fakeLog = sandbox.stub();
     lib.setLog(fakeLog);
     sandbox.stub(composer, 'resolveComponentReferences');
@@ -43,12 +44,13 @@ describe(_.startCase(filename), function () {
     const fn = lib[this.title];
 
     it('will try to invoke each plugin', function () {
-      lib.plugins = [plugin];
+      lib.plugins = [plugin, pluginWithMetadata];
       plugin.returns(true);
 
       return fn()
         .then(() => {
           sinon.assert.calledOnce(plugin);
+          sinon.assert.calledOnce(pluginWithMetadata.init);
         });
     });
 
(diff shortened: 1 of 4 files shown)


lib/services/plugins.test.js
```diff
@@ -43,23 +44,59 @@ describe(_.startCase(filename), function () {
...
       return fn()
         .then(() => {
-          sinon.assert.calledWith(fakeLog, 'warn', 'Plugin is not a function');
+          sinon.assert.calledWith(fakeLog, 'warn', 'Plugin is not a function or does not have an init function');
         });
     });
   });
+
+  describe('getWhitelistFromPlugins', function () {
+    it('returns an array of all the plugin\'s whitelist combined', function () {
```
manuelurenah: Can we cache the function that's being tested so we don't have to call `lib.getWhitelistFromPlugins` in every test case?
```js
describe('getWhitelistFromPlugins', function () {
  const fn = lib[this.title];

  it('returns an array of all the plugin\'s whitelist combined', function () {
    // ...

    expect(fn(plugins)).to.eql(['/_test']);
  });
});
```

Do this for every test suite

lib/services/plugins.test.js
```diff
@@ -43,23 +44,59 @@ describe(_.startCase(filename), function () {
...
+
+      plugin.returns(true);
+      lib.getWhitelistFromPlugins(plugins);
+
+      expect(lib.getWhitelistFromPlugins(plugins)).to.be.empty;
+    });
+  });
+
+  describe('skipAuth', function () {
+    it('sets skipAuth to true if the request url is in the whitelist', function () {
```
manuelurenah: Same suggestion as before

PedroAlbR:
Closing this since the approach is not what we are interested in at the moment.

PR #656, opened 2019-04-15
Pass down locals in component bootstrap

### LINKED ISSUE #653: Locals Not Available On Startup Bootstrap
## Description

The `locals` object is not available during [startup bootstrapping](https://docs.clayplatform.com/docs/glossary#bootstrap-action).

## Steps to Reproduce

1. Build a component whose model.js file references the `locals` object in the `save` method
2. Restart the server
3. Notice that `locals` is undefined

## Expected Behavior

We should pass in a `locals` object formed in the same structure as if a request was coming into the component directly through the `_components/<NAME>` route

## Additional Context

Here is where the component saves are being generated: https://github.com/clay/amphora/blob/master/lib/bootstrap.js#L217-L222

manuelurenah:
### Description

* Fixes issue where locals were not being passed down when bootstrapping a component
* Closes #653

### DIFF
--- a/lib/bootstrap.js
+++ b/lib/bootstrap.js
@@ -10,7 +10,8 @@ const path = require('path'),
   siteService = require('./services/sites'),
   references = require('./services/references'),
   highland = require('highland'),
-  { encode } = require('./services/buffer');
+  { encode } = require('./services/buffer'),
+  locals = require('./locals');
 var log = require('./services/logger').setup({
     file: __filename,
     action: 'bootstrap'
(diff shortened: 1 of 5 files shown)


lib/bootstrap.js
```diff
@@ -207,7 +207,8 @@ function applyPrefix(data, site) {
 function load(data, site) {
   let promiseComponentsOps, promisePagesOps, promiseUrisOps, promiseListsOps, promiseUsersOps,
     prefix = site.prefix,
-    compData = applyPrefix(data, site);
+    compData = applyPrefix(data, site),
+    locals = { site, components: files.getComponents() };
```
jonwinton: I'm wondering if we need to stub some of the other properties available on the object by default. 

- `locals.url`
- `locals.query`
- `locals.params`
- `locals.edit`

https://docs.clayplatform.com/docs/glossary#locals

Those are just some that are nice to not have to check for when writing model.js files

PR #657, opened 2019-05-01
Allow plugins to opt-in/out of authentication

PedroAlbR:
### Description

- Passes a function (`useAuth`) to each plugin.

### Plugin example

```js
const express = require('express');

module.exports = (router, _pluginDBAdapter, _publish, _sites, _site, useAuth) => {
  const pluginRouter = express.Router();

  useAuth(pluginRouter);
  pluginRouter.use((req, res) => res.json({ method: req.method }));
  router.use('/_test', pluginRouter);
}
```

### Related
https://github.com/clay/amphora-auth/pull/19

### DIFF
--- a/lib/routes.test.js
+++ b/lib/routes.test.js
@@ -27,6 +27,7 @@ describe(_.startCase(filename), function () {
     fakeLog = sandbox.stub();
     fakeAuth = sandbox.stub();
     fakeAuth.addRoutes = sandbox.stub();
+    fakeAuth.useAuth = sandbox.stub();
 
     lib.setLog(fakeLog);
     lib.setamphoraAuth(fakeAuth);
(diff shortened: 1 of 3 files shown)


lib/routes.test.js
```diff
@@ -27,6 +27,7 @@ describe(_.startCase(filename), function () {
     fakeLog = sandbox.stub();
     fakeAuth = sandbox.stub();
     fakeAuth.addRoutes = sandbox.stub();
+    fakeAuth.useAuth = sandbox.stub();
```
manuelurenah: Are we asserting that `useAuth` is being called? If not, then there should be a test case which makes the assertion to make sure the method is being called with the expected arguments

PedroAlbR:
[Amphora Plugins Auth Requirement](https://trello.com/c/cvUJKrMc/122-amphora-plugins-auth-requirement)

Issue #633, opened 2019-02-22
Archive Event Not Added To History


jonwinton:
## Description

When a page is archived it's not being recorded in the page history

## Steps to Reproduce

1. Make a new page
2. Archive it
3. View the page history at the `/meta` endpoint

## Expected Behavior

The archived event is there

## Additional Context

Kiln might need to be updated to handle rendering this event.

PR #659, opened 2019-05-14
Updates patchMeta method to add archive events to history

manuelurenah:
### Description

* Modifies `patchMeta` method in metadata controller in order to be able to know when to add archive events in the metadata history
* This closes issue #633 

### QA

* Do `npm link` to `clay-kiln` in branch `update-archive`
* Archive/Unarchive any article
* Go to page metadata in `<page_uri>/meta`
* Events should be logged inside `history` array

### DIFF
--- a/lib/services/metadata.js
+++ b/lib/services/metadata.js
@@ -224,12 +224,27 @@ function putMeta(uri, data) {
  *
  * @param {String} uri
  * @param {Object} data
+ * @param {Object} locals
  * @returns {Promise}
  */
-function patchMeta(uri, data) {
-  const id = replaceVersion(uri.replace('/meta', ''));
+function patchMeta(uri = '', data = {}, locals = {}) {
+  const id = replaceVersion(uri.replace('/meta', '')),
+    update = {
+      updateTime: data.updateTime,
+      history: data.history,
+      archived: data.archived,
+      users: data.users
+    };
 
-  return db.patchMeta(id, data)
+  if (data.shouldPatchArchive) {
+    const { user } = locals,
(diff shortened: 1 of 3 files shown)


lib/services/metadata.test.js
```diff
@@ -69,13 +69,53 @@ describe(_.startCase(filename), function () {
       db.patchMeta.returns(Promise.resolve());
       db.getMeta.returns(Promise.resolve());
 
-      return fn('domain.com/_pages/id/meta', {})
+      return fn('domain.com/_pages/id/meta', {}, {})
```
fxisco: you set the default for the 2nd and 3rd argument in the function, they can be removed

lib/services/metadata.js
```diff
@@ -224,11 +224,21 @@ function putMeta(uri, data) {
...
  * @returns {Promise}
  */
-function patchMeta(uri, data) {
+function patchMeta(uri = '', data = {}, locals = {}) {
   const id = replaceVersion(uri.replace('/meta', ''));
 
+  if (data.shouldPatchArchive) {
+    const { user } = locals,
+      action = data.archived ? 'archive' : 'unarchive',
+      historyEntry = { action, timestamp: new Date().toISOString(), users: [userOrRobot(user)] };
```
fxisco: You could have the logic to create a historyEntry separately. This will improve testing.

lib/services/metadata.test.js
manuelurenah: I sent these params in here so anyone who have a look at the tests knows that it receives more than one param

PR #663, opened 2019-05-23
Update docs page color

oscarpolanco:
Update page color

# Steps for testing the site locally:
- Go to the `website` folder
- Install all dependencies using `npm install`
- Start your local server using `npm run start`
- Test locally at [localhost](http://localhost:3000/amphora/)
- Compare with the current docs [page](https://docs.clayplatform.com/amphora/)

### DIFF
--- a/docs/introduction.md
+++ b/docs/introduction.md
@@ -41,7 +41,9 @@ npm install --save amphora
 
 ## Usage
 
-Clay separates concerns into two main areas: components and sites. Create two new directories in your project:
+Clay separates concerns into two main areas: components and sites. 
+
+First, create two new directories in your project:
 
 ```text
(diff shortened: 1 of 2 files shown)


docs/introduction.md
```diff
@@ -41,7 +41,9 @@ npm install --save amphora
 
 ## Usage
 
-Clay separates concerns into two main areas: components and sites. Create two new directories in your project:
+Clay separates concerns into two main areas: components and sites.
```
Shilario:
```suggestion
Clay separate concerns into two main areas: components and sites. 
```

manuelurenah: Nice suggestion Steph, but the correct way for singular would be as it was before. So, it is good as it is right now

PR #664, opened 2019-05-28
[WIP] Use semaphore from the storage module

### LINKED ISSUE #606: Startup Semaphore
[Quick semaphore link](https://en.wikipedia.org/wiki/Semaphore_(programming))

# Problem
In a production environment there are generally multiple Clay instances running at one time, whether it be multiple servers/containers or even a [cluster](https://nodejs.org/api/cluster.html) on a single instance. At startup time Clay bootstraps data for components/sites, something we only need/want to run on one individual instance in a cluster/deployment group.

# Proposal
Add semaphore support to the storage API, allowing whichever storage module is chosen to dictate how the data is stored. The data we might send across:

- Which node has begun bootstrapping. Could be determined with the `os` module using a hostname/pid combination to generate the key.
- Which deploy is being handled. Can we rely on the package.json version of the clay instance? Multiple environments are not always going to see this increment
- The state of the startup process: Not started, running, finished, failed? 0, 1, 2, 3? 

# Questions
- What happens if the startup process fails and needs to be restarted? 
- Should plugins be able to use the same semaphore data to determine how they startup? For example, amphora-search does not ALWAYS need to check for the presence of indices on every Clay instance.

PedroAlbR:
Closes #606

### DIFF
--- a/lib/setup.js
+++ b/lib/setup.js
@@ -64,7 +64,10 @@ module.exports = function (options = {}) {
     .then(() => db.registerStorage(storage))
     .then(() => amphoraPlugins.registerPlugins(plugins))
     .then(() => { router = routes(app, providers, sessionStore); })
-    .then(() => internalBootstrap(bootstrap))
+
+    // TODO: Must make this optional
+    .then(() => db.applyLock('bootstrap', () => internalBootstrap(bootstrap)))
+
     .then(() => router);
 };
 
--- a/lib/services/db.js
+++ b/lib/services/db.js
@@ -130,6 +130,7 @@ module.exports.registerStorage = registerStorage;
 module.exports.put;
 module.exports.get;
 module.exports.del;
+module.exports.applyLock;
 module.exports.batch;
 module.exports.putMeta;
 module.exports.patchMeta;


amelvisfranco:
[Semaphora](https://trello.com/c/Wd9zi43D/92-semaphora)

lib/setup.js
```diff
@@ -64,7 +64,10 @@ module.exports = function (options = {}) {
     .then(() => db.registerStorage(storage))
     .then(() => amphoraPlugins.registerPlugins(plugins))
     .then(() => { router = routes(app, providers, sessionStore); })
-    .then(() => internalBootstrap(bootstrap))
+    .then(() =>
+      db.applyLock
+        ? db.applyLock('bootstrap', internalBootstrap)
```
james-owen: Looks good, but I'd like to get some clarity on this pattern.

fxisco: should action name be uppercased?

PedroAlbR: This was a suggestion to avoid breaking changes:
If the storage module has an `applyLock` method, we use it.
If it doesn't, we execute the bootstrap as before (using the `bootstrap` boolean that comes from the configuration)

PR #668, opened 2019-06-11
Update the docs and website README

oscarpolanco:
Update the `README` of the `docs` and `website` directory with the info and step to work with the documentation

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,31 @@
+# Clay Documentation
+
+On this directory, we have the pre-release documentation written using markdown. This file has the `.md` extension and as we mentioned before, these are pre-release files if you don't add a documentation version yet.
+
+## Markdown options
+
+These files have the following structure at the beginning of the file:
+
+```
+---
+id: idOftheFile
+title: Title of your file
+sidebar_label: Title of the sidebar
+---
+
+My new content here..
+```
+
+The previous example has a couple of the most important part to identify the content that you wish to add to your documentation. Here is a more detailed explanation  on the [Markdown feature](https://docusaurus.io/docs/en/doc-markdown) that you can use.
+
+## Test documentation page
+
+We have a local server that we use to test your current docs content. Here are the steps on how to set the [local server](https://github.com/clay/amphora/tree/master/website#run-the-local-server). Normally you will have access on your browser with a url like this:
+
+`http://localhost:3000/amphora/docs/idOfTheDocs`
+
+But if you already have a [version](https://github.com/clay/amphora/tree/master/website#versioning-the-code) on your docs you will need to add `next` on your url
+
+`http://localhost:3000/amphora/docs/next/idOfTheDocs`
+
+[Here](https://docusaurus.io/docs/en/versioning) are more insights about what to do when you already have documentation with version.
--- a/website/README.md
+++ b/website/README.md
@@ -7,7 +7,9 @@ You will need to have `Node >= 8.x` and `Yarn >= 1.5`.
 ```sh
 $ yarn
 ```
+
 or
+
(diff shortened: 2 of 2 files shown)


docs/README.md
```diff
@@ -0,0 +1,31 @@
+# Clay Documentation
+
+On this directory, we have the pre-release documentation written using markdown. This file has the `.md` extension and as we mentioned before these are pre-release files if you don't add a documentation version yet.
```
manuelurenah:
```suggestion
On this directory, we have the pre-release documentation written using markdown. This file has the `.md` extension and as we mentioned before, these are pre-release files if you don't add a documentation version yet.
```

docs/README.md
```diff
@@ -0,0 +1,31 @@
...
+---
+id: idOftheFile
+title: Title of your file
+sidebar_label: Title of the sidebar
+---
+
+My new content here..
+```
+
+The previews example has a couple of the most important part to identify the content that you wish to add to your documentation. Here is a more detail explanation  on the [Markdown feature](https://docusaurus.io/docs/en/doc-markdown) that you can use.
```
manuelurenah:
```suggestion
The previous example has a couple of the most important part to identify the content that you wish to add to your documentation. Here is a more detailed explanation on the [Markdown feature](https://docusaurus.io/docs/en/doc-markdown) that you can use.
```

docs/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+We have a local server that we use to test your current docs content. Here are the steps on how to set the [local server](https://github.com/clay/amphora/tree/master/website#run-the-local-server). Normally you will have access on your browser with a url like this:
+
+`http://localhost:3000/amphora/docs/idOfTheDocs`
+
+But if you already have a [version](https://github.com/clay/amphora/tree/master/website#versioning-the-code) on your docs you will need to add `next` on your url
+
+`http://localhost:3000/amphora/docs/next/idOfTheDocs`
+
+[Here](https://docusaurus.io/docs/en/versioning) are more inside when you have you already have documentation with version.
```
manuelurenah:
```suggestion
[Here](https://docusaurus.io/docs/en/versioning) are more insights about what to do when you already have documentation with version.
```

website/README.md
```diff
@@ -55,21 +59,61 @@ root-directory
...
 ```
-$ yarn build
+
+or
+
+```sh
+npm run examples versions
+```
+
+Then you proceed to create a version of the docs that you need in this case need to match the version of your current `package.json`. To do this we will use the following command followed the number of the version that you want.
```
manuelurenah:
```suggestion
Then, you proceed to create a version of the docs that you need. In this case, it needs to match the version of your current `package.json`. To do this we will use the following command followed by the version number that you want.
```

website/README.md
```diff
@@ -55,21 +59,61 @@ root-directory
...
+$ yarn version X.X.X
 ```
+
 or
+
+```sh
+$ npm version X.X.X
+```
+
+This will create all the files need it to handle the version that you just build but this will be necessary only in 2 cases:
```
manuelurenah:
```suggestion
This will create all the files needed to handle the version that you just built, but this will be necessary only in 2 cases:
```

website/README.md
```diff
@@ -55,21 +59,61 @@ root-directory
...
+$ yarn build
 ```
+
+or
+
+```sh
 $ npm run build
 ```
 
 Then you just need to deploy the static files generated. These files will be pushed to the `gh-page` branch.
```
manuelurenah:
```suggestion
Then, deploy the static files generated. These files will be pushed to the `gh-page` branch.
```

PR #669, opened 2019-06-24
Add the missing route information

oscarpolanco:
Ticket: https://app.zenhub.com/workspaces/clay-repos-596524da4a8a0769f2f03175/issues/clay/amphora/499

Add missing `route` information that targets the `scheduling` part of the `page` ticket

# Steps for testing the site locally:
- Go to the `website` folder
- Install all dependencies using `npm install`
- Start your local server using `npm run start`
- Test locally the current routes page at [localhost pre-release](http://localhost:3001/amphora/docs/next/routes)
- Test locally the new page at [localhost](http://localhost:3001/amphora/docs/next/route)
- Check that the left sidebar has a new route division with 2 pages at [localhost](http://localhost:3001/amphora/docs/next/routes)

### DIFF
--- a/docs/routes.md
+++ b/docs/routes.md
@@ -1,0 +1,268 @@
+---
+id: route
+title: Technical information
+sidebar_label: Technical information
+---
+
+This is a technical document about how routing in Amphora is performed. It also informs any further technical decisions about the growth of the library and contains many ideas that we're adhering to for consistency.
+
+## Components
+
+### Overview
+
+```
+/components
+/components/:name
+/components/:name@:version
+/components/:name.html
+/components/:name.json
+/components/:name@:version.html
+/components/:name@:version.json
+/components/:name/instances
+/components/:name/instances/:id
+/components/:name/intsances/:id.html
+/components/:name/intsances/:id.json
+/components/:name/instances/:id@:version
+/components/:name/instances/:id@:version.html
+/components/:name/instances/:id@:version.json
+```
+
+### List of components
+`GET /components` will return a list of the available known components. This will be all the components defined in the `/components` folder of the current application, and also any components installed through the node package manager (npm).  
+
+Example: `GET /components`
+```json
+[
+  "text",
+  "image",
+  "paragraph",
+  "article"
+]
+```
+
+`GET /components/:name/instances` will return a list of the available instances within a particular component type.
+
+Example: `GET /components/text/instances`
+```json
+[
+  "/components/text/instances/abc",
+  "/components/text/instances/def",
+  "/components/text/instances/def@published"
+]
+```
+
+### Component Data
+
+You can grab the default data for a component, data for a specific instance, or even data for a specific version (of an instance). GETs and PUTs to these endpoints behave as you would expect from a RESTful API (they return what you send them). GETs and PUTs to the `.json` and `.html` extensions also work, but they'll return the composed (i.e. a component and its children) JSON and HTML, respectively.
+
+### Modifying components
+`GET /components/:name[/instances/:id][@:version[.:extension]]` will return a component. The data is based on the extension (.html, .json, .yaml), or the `Accepts` header of the request.
+
+Requesting data without a version will return the latest saved version. Some versions have special rules (see [Propagating Versions](#propagating-versions)), but any other version name can be used to specially tag any particular version.
+
+`PUT /components/:name[/instances/:id][@version]` will save the data such that `GET`ing the same URI will return exactly what was put there.
+
+### Component Logic
+
+Components may have pre-render and pre-save hooks, which do logic before rendering and saving, respectively. To add these, create a `model.js` file in the component's folder, and export `render` and/or `save` methods. Both of these methods should return objects with the component data, which may be wrapped in promises.
+
+_Note:_ It is much better to do component logic when saving rather than every time it wants to render.
+
+```js
+module.exports.render = function (ref, data, locals) {
+  // do logic
+  return data; // will be sent to the template
+}
+
+module.exports.save = function (ref, data, locals) {
+  // do logic
+  return data; // will be sent to the database
+}
+```
+
+You may pass `componenthooks=false` as a query param when doing API calls if you want to completely bypass the `model.js` (e.g. if you've already run through the logic client-side).
+
+### Legacy Server Logic
+
+This feature is deprecated as of amphora v2.11.0 and will be removed in the next major version (it is supplanted by the isomorphic `model.js` files). Legacy server-side logic lives in a `server.js` or `index.js` file in the component's folder, and has a few differences from the `model.js`:
+
+* The default exported function (run on `GET`) does not automatically receive data from the database and must fetch it manually
+* The `put()` should return a database operation or array of operations (which may be wrapped in promises), rather than the component data itself
+* These files are _only_ run server-side, and will not be optimized to work in kiln
+
+```js
+module.exports = function (ref, locals) {
+  //return Promise with data
+  return Promise.resolve({"hey": "hey"});
+};
+module.exports.put = function (ref, data) {
+  //return Promise with operations to be performed in a batch
+  return Promise.resolve([{
+    type: 'put',
+    key: '/components/text',
+    value:'{"hey": "hey"}'}
+  ]);
+};
+```
+
+## Pages
+
+### Overview
+```
+/pages
+/pages/:name
+/pages/:name.html
+/pages/:name.json
+/pages/:name@:version
+/pages/:name@:version.html
+/pages/:name@:version.json
+```
+
+Pages consist of a layout and a list of areas. Each area defined in a page maps to an area in the layout template. For example:
+```json
+{
+  "layout": "/components/feature-layout",
+  "center": ["/components/article/instances/3vf3"],
+  "side": [
+    "/components/share",
+    "/components/newsletter"
+  ]
+}
+```
+The layout component in this example has two areas: center and side. When the page is displayed, the components listed here will be placed into the matching areas in the layout.
+
+### Page Data
+
+`GET`s and `PUT`s to pages work similarly to components. API calls without an extension will update/return data for that page, while the `.json` and `.html` extensions will return the composed (page and layout, and their child components) JSON and HTML, respectively.
+
+### Publishing
+
+Publishing a page has a special convenience behavior where all components referenced by the page will also be published. (Example: https://github.com/nymag/amphora/pull/78)
+
+## URIs
+
+```
+/uris/<base64(:path)>
+```
+
+A URI is used to redirect some slug or URI to another page or component. They can also redirect to other URIs (establishing a 301 redirect), or several URIs can point to the same resource. URIs are stored as `Base64`, so:
+
+- `example.com` is `/uris/ZXhhbXBsZS5jb20=` => `/pages/jdskla@published`
+- `example.com/other/` is `/uris/ZXhhbXBsZS5jb20vb3RoZXI=` => `/pages/4revd3s@published`
+
+A URI is assumed to be pointing at the `@published` version if another version is not provided. Therefore, only published content or specially tagged versions can be publicly exposed through URIs.
+
+## Lists
+
+```
+/lists
+/lists/:id
+```
+
+Lists are a _temporary_ solution until search functionality is discussed. It is currently used for the tags component and the autocomplete behavior as a temporary place to store lists of information. It can store any list of data on a `PUT`, and return back the same list of data with a `GET`.
+
+## Users
+
+### Overview
+```
+/users
+/users/:id
+```
+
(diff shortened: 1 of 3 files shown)


docs/routes.md
```diff
@@ -0,0 +1,268 @@
...
+```
+
+### Component Data
+
+You can grab the default data for a component, data for a specific instance, or even data for a specific version (of an instance). GETs and PUTs to these endpoints behave as you would expect from a RESTful API (they return what you send them). GETs and PUTs to the `.json` and `.html` extensions also work, but they'll return the composed (i.e. a component and its children) JSON and HTML, respectively.
+
+### Modifying components
+`GET /components/:name[/instances/:id][@:version[.:extension]]` will return a component.  The form the data is based on the extension (.html, .json, .yaml), or the Accepts header of the request.  
+
+Requesting data without a version will return the latest saved version.  Some versions have special rules (see [Propagating Versions](#propagating-versions)), but any other version name can be used to specially tag any particular version.
```
manuelurenah:
```suggestion
Requesting data without a version will return the latest saved version. Some versions have special rules (see [Propagating Versions](#propagating-versions)), but any other version name can be used to specially tag any particular version.
```

docs/routes.md
```diff
@@ -0,0 +1,268 @@
...
+}
+```
+
+You may pass `componenthooks=false` as a query param when doing API calls if you want to completely bypass the `model.js` (e.g. if you've already run through the logic client-side).
+
+### Legacy Server Logic
+
+This feature is deprecated as of amphora v2.11.0 and will be removed in the next major version (it is supplanted by the isomorphic `model.js` files). Legacy server-side logic lives in a `server.js` or `index.js` file in the component's folder, and has a few differences from the `model.js`:
+
+* the default exported function (run on `GET`) does not automatically receive data from the database and must fetch it manually
```
manuelurenah:
```suggestion
* The default exported function (run on `GET`) does not automatically receive data from the database and must fetch it manually
```

docs/routes.md
```diff
@@ -0,0 +1,268 @@
...
+* the `put()` should return a database operation or array of operations (which may be wrapped in promises), rather than the component data itself
```
manuelurenah:
```suggestion
* The `put()` should return a database operation or array of operations (which may be wrapped in promises), rather than the component data itself
```

docs/routes.md
```diff
@@ -0,0 +1,268 @@
...
+* these files are _only_ run server-side, and will not be optimized to work in kiln
```
manuelurenah:
```suggestion
* These files are _only_ run server-side, and will not be optimized to work in kiln
```

docs/routes.md
```diff
@@ -0,0 +1,268 @@
...
+/pages
+/pages/:name
+/pages/:name.html
+/pages/:name.json
+/pages/:name@:version
+/pages/:name@:version.html
+/pages/:name@:version.json
+```
+
+Pages consist of a layout and a list of areas.  Each area defined in an page maps to a area in the layout template.  For example:
```
manuelurenah:
```suggestion
Pages consist of a layout and a list of areas. Each area defined in a page maps to an area in the layout template. For example:
```

docs/routes.md
```diff
@@ -0,0 +1,268 @@
...
+{
+  "layout": "/components/feature-layout",
+  "center": ["/components/article/instances/3vf3"],
+  "side": [
+    "/components/share",
+    "/components/newsletter"
+  ]
+}
+```
+The layout component in this example has two areas: center and side.  When the page is displayed, the components listed here will be placed into the matching areas in the layout.
```
manuelurenah:
```suggestion
The layout component in this example has two areas: center and side. When the page is displayed, the components listed here will be placed into the matching areas in the layout.
```

docs/routes.md
```diff
@@ -0,0 +1,268 @@
...
+
+### Page Data
+
+GETs and PUTs to pages work similarly to components. API calls without an extension will update/return data for that page, while the `.json` and `.html` extensions will return the composed (page and layout, and their child components) JSON and HTML, respectively.
```
manuelurenah:
```suggestion
`GET`s and `PUT`s to pages work similarly to components. API calls without an extension will update/return data for that page, while the `.json` and `.html` extensions will return the composed (page and layout, and their child components) JSON and HTML, respectively.
```

docs/routes.md
```diff
@@ -0,0 +1,268 @@
...
+
+### Publishing
+
+Publishing a page has a special convenience behavior where all components referenced by the page will also be published.  (Example: https://github.com/nymag/amphora/pull/78)
```
manuelurenah:
```suggestion
Publishing a page has a special convenience behavior where all components referenced by the page will also be published. (Example: https://github.com/nymag/amphora/pull/78)
```

docs/routes.md
```diff
@@ -0,0 +1,268 @@
...
+
+## URIs
+
+```
+/uris/<base64(:path)>
+```
+
+A URI is used to redirect some slug or URI to another page or component.  They can also redirect to other uris (establishing a 301 redirect), or several uris can point to the same resource.  URIs are stored as Base64, so:
```
manuelurenah:
```suggestion
A URI is used to redirect some slug or URI to another page or component. They can also redirect to other URIs (establishing a 301 redirect), or several URIs can point to the same resource. URIs are stored as `Base64`, so:
```

docs/routes.md
```diff
@@ -0,0 +1,268 @@
...
+
+- `example.com` is `/uris/ZXhhbXBsZS5jb20=` => `/pages/jdskla@published`
+- `example.com/other/` is `/uris/ZXhhbXBsZS5jb20vb3RoZXI=` => `/pages/4revd3s@published`
+
+A URI is assumed to be pointing at the `@published` version if another version is not provided.  Therefore, only published content or specially tagged versions can be publicly exposed through URIs.
```
manuelurenah:
```suggestion
A URI is assumed to be pointing at the `@published` version if another version is not provided. Therefore, only published content or specially tagged versions can be publicly exposed through URIs.
```

docs/routes.md
```diff
@@ -0,0 +1,268 @@
...
+
+## Lists
+
+```
+/lists
+/lists/:id
+```
+
+Lists are a _temporary_ solution until search functionality is discussed.  It is currently used for the tags component and the autocomplete behavior as a temporary place to store lists of information.  It can store any list of data on a `PUT`, and return back the same list of data with a `GET`.
```
manuelurenah:
```suggestion
Lists are a _temporary_ solution until search functionality is discussed. It is currently used for the tags component and the autocomplete behavior as a temporary place to store lists of information. It can store any list of data on a `PUT`, and return back the same list of data with a `GET`.
```

docs/routes.md
```diff
@@ -0,0 +1,268 @@
...
+
+## Users
+
+### Overview
+```
+/users
+/users/:id
+```
+
+Each user has a `username` and `provider`, which determines how they authenticate over oauth. Users can also have other data, including `name`, `imageUrl`, and `title`.
```
manuelurenah:
```suggestion
Each user has a `username` and `provider`, which determines how they authenticate over OAuth. Users can also have other data, including `name`, `imageUrl`, and `title`.
```

docs/routes.md
```diff
@@ -0,0 +1,268 @@
...
+
+## Schedule
+
+```
+/schedule
+/schedule/:id
+```
+
+Schedule is a list of pages or components that will be published in the future.  Items in this list will be published when the time passes.  Each item has two properties: `at`, which is a UNIX timestamp, and the `publish` property that is a reference to what will be published after that time.
```
manuelurenah:
```suggestion
Schedule is a list of pages or components that will be published in the future. Items in this list will be published when the time passes. Each item has two properties: `at`, which is a UNIX timestamp, and the `publish` property that is a reference to what will be published after that time.
```

docs/routes.md
```diff
@@ -0,0 +1,268 @@
...
+
+When an item is scheduled, a `@scheduled` version is created on the item that will be published.  When the item is eventually published, the `@scheduled` version will be removed.  The `@scheduled` version is a reference to the actual schedule URI that can be deleted to cancel the scheduled publication.
```
manuelurenah:
```suggestion
When an item is scheduled, a `@scheduled` version is created on the item that will be published. When the item is eventually published, the `@scheduled` version will be removed. The `@scheduled` version is a reference to the actual schedule URI that can be deleted to cancel the scheduled publication.
```

docs/routes.md
```diff
@@ -0,0 +1,268 @@
...
+### Add a scheduled item
+
+`POST /schedule` will add an item to be published in the future. The format of the item should be of
+```json
+{
+  "at": 44392893402093,
+  "publish": "<uri to be published>"
+}
+```
+where the `at` is a UNIX timestamp and the `publish` is a ref to a page or component.  On success, it will return a 201 with a reference to the schedule URI that was created.
```
manuelurenah:
```suggestion
where the `at` is a UNIX timestamp and the `publish` is a ref to a page or component. On success, it will return a 201 with a reference to the schedule URI that was created.
```

docs/routes.md
```diff
@@ -0,0 +1,268 @@
...
+
+
+### Remove a scheduled item
+
+`DELETE /schedule/:some-id` will remove the scheduling of a publish in the future.  On success, it will return a 200 with
```
manuelurenah:
```suggestion
`DELETE /schedule/:some-id` will remove the scheduling of a publish in the future. On success, it will return a 200 with
```

docs/routes.md
```diff
@@ -0,0 +1,268 @@
...
+
+Some versions have a special behavior called _propagating_, which any reference within their data is changed to be the same version as itself.
+
+The current propagating versions are:
+* published
+* latest
+
+## RESTful API
+
+We try to follow the [REST](https://en.wikipedia.org/wiki/Representational_state_transfer) standard when making decisions about changes to the API.  In particular, we are adhering to
```
manuelurenah:
```suggestion
We try to follow the [REST](https://en.wikipedia.org/wiki/Representational_state_transfer) standard when making decisions about changes to the API. In particular, we are adhering to
```

docs/routes.md
```diff
@@ -0,0 +1,268 @@
...
+
+1.  _Individual resources are identified by URI._  Therefore, there is no need to return an id as part of the data of a request if that id was used to fetch the data.  In many environments (production, development, staging) hostnames can be dynamic, so we can optionally omit the hostname and assume that the resource is available at the same host as a referring resource.
```
manuelurenah:
```suggestion
1.  _Individual resources are identified by URI._  Therefore, there is no need to return an `id` as part of the data of a request if that `id` was used to fetch the data. In many environments (production, development, staging) hostnames can be dynamic, so we can optionally omit the hostname and assume that the resource is available at the same host as a referring resource.
```

docs/routes.md
```diff
@@ -0,0 +1,268 @@
...
+
+    Currently, we link to child resources with `{ _ref: 'domain.com/full/resource/uri' }`. There is no assumption that the client knows about these child resources implicitly.  More importantly, we don't expect the client to know about where resources are, their types, or that they should be linked together in any particular way, so we always reference to linked resources with a full URI.
+
+    For more information about the design decisions of `_ref`, see https://github.com/nymag/amphora/issues/108
+
+### Namespacing
+
+It might be a good idea to namespace this API behind versions, such as `/v1/` or `/v2/`, or `/services/`, or even `/clay/`.  This can be done by adjusting the hostname, or through a reverse proxy like Varnish.  Putting different versions behind Varnish might be particularly useful so that previous versions of an application can be run at the same time as newer versions to service old and new clients until previous versions can be deprecated.
```
manuelurenah:
```suggestion
It might be a good idea to namespace this API behind versions, such as `/v1/` or `/v2/`, or `/services/`, or even `/clay/`. This can be done by adjusting the hostname, or through a reverse proxy like Varnish. Putting different versions behind Varnish might be particularly useful so that previous versions of an application can be run at the same time as newer versions to service old and new clients until previous versions can be deprecated.
```

docs/routes.md
```diff
@@ -0,0 +1,268 @@
...
+
+## Errors
+
+### Ordering of HTTP Errors
+
+The API must follow a specific ordering for the priority of errors to return.  Even though _multiple_ things might be
```
manuelurenah:
```suggestion
The API must follow a specific ordering for the priority of errors to return. Even though _multiple_ things might be
```

docs/routes.md
```diff
@@ -0,0 +1,268 @@
...
+wrong with a request, consistent testability requires that the type of error returned for any particular request should
+not change without deliberate thought.  The current order of errors are:
```
manuelurenah: This line can be put together with the one above
```suggestion
not change without deliberate thought. The current order of errors are:
```

PR #671, opened 2019-07-11
User's data normalization

jjpaulino:
Solving an issue where the users data in the meta of the page and the `_users` endpoint wasn't synced.

We'll only handle the properties that shouldn't change and request for the changeable information (like the name or picture) everytime we need. 

This is going to work in conjunction with a clay-kiln, amphora-auth and amphora-search change


jjpaulino:
[Denormalize page-list users](https://trello.com/c/R2pJxeCP/25-denormalize-page-list-users)

PR #674, opened 2019-07-31
Subsite support

jpope19:
dependencies:
- https://github.com/clay/amphora-auth/pull/26

adds subsite support. allows for the following file structure inside of sites:

```
sites
    - site-name
        - subsites
            - subsite-name
                - media
                index.js
                bootstrap.yaml
                config.yaml
    index.js
    bootstrap.yaml
    config.yaml
```

required config for subsites:
- protocol
- path
- host
- port
- assetPath

### DIFF
--- a/lib/bootstrap.js
+++ b/lib/bootstrap.js
@@ -297,6 +297,8 @@ function bootstrapSite(site) {
     .then(function () {
       if (site.subsite) {
         return bootstrapPath(site.dir, site);
+      } else {
+        return site;
       }
     })
     .catch(function () {
--- a/lib/bootstrap.js
+++ b/lib/bootstrap.js
@@ -283,11 +283,12 @@ function bootstrapError(component, e) {
 /**
  * Bootstraps a site, and if it fails, falls back to the parent sites' bootstrap (for subsites)
  *
- * @param {Object} site 
+ * @param {Object} site
+ * @return {Promise}
  */
 function bootstrapSite(site) {
   return bootstrapPath(site.dir, site)
-    .catch(function(e) {
+    .catch(function (e) {
       if (site.subsite) {
         log('info', `No bootstrap file found for site: ${site.slug} at ${site.dir} so checking for parent site.`);
         const parent = siteService.getParentSite(site.slug);
--- a/lib/bootstrap.js
+++ b/lib/bootstrap.js
@@ -292,17 +292,11 @@ function bootstrapSite(site) {
 
   // using module.exports for test
   return module.exports.bootstrapPath(parent.dir, site)
-    .catch(function () {
+    .catch(() => {
       log('debug', `No bootstrap file found for parent site: ${site.slug}`);
     })
-    .then(function () {
-      if (site.subsite) {
-        return module.exports.bootstrapPath(site.dir, site);
-      } else {
-        return site;
-      }
-    })
-    .catch(function () {
+    .then(() => site.subsite ? module.exports.bootstrapPath(site.dir, site) : site)
+    .catch(() => {
       log('debug', `No bootstrap file found for subsite: ${site.slug} ${site.subsite}`);
       return site;
     });
--- a/lib/bootstrap.js
+++ b/lib/bootstrap.js
@@ -281,6 +281,25 @@ function bootstrapError(component, e) {
 }
 
 /**
+ * Bootstraps a site, and if it fails, falls back to the parent sites' bootstrap (for subsites)
+ *
+ * @param {Object} site 
+ */
+function bootstrapSite(site) {
+  return bootstrapPath(site.dir, site)
+    .catch(function(e) {
+      if (site.subsite) {
+        log('info', `No bootstrap file found for site: ${site.slug} at ${site.dir} so checking for parent site.`);
+        const parent = siteService.getParentSite(site.slug);
+
+        return bootstrapPath(parent.dir, site);
+      } else {
+        throw e;
+      }
+    });
+}
+
+/**
  * Bootstrap all sites and the individual
  * bootstrap files in each component's directory
  *
@@ -305,7 +324,7 @@ function bootrapSitesAndComponents() {
     })
     .flatMap(site => {
       return highland(
-        bootstrapPath(site.dir, site)
+        bootstrapSite(site)
           .catch(function () {
             log('debug', `No bootstrap file found for site: ${site.slug}`);
             return site;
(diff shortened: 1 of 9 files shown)


lib/bootstrap.js
```diff
@@ -280,6 +280,34 @@ function bootstrapError(component, e) {
...
+ * @param {Object} site
+ * @return {Promise}
+ */
+function bootstrapSite(site) {
+  // fetches itself if it is the main site
+  const parent = siteService.getParentSite(site.slug);
+
+  // using module.exports for test
+  return module.exports.bootstrapPath(parent.dir, site)
+    .catch(function () {
```
manuelurenah: Let's replace this with arrow function
```suggestion
    .catch(() => {
```

lib/bootstrap.test.js
```diff
@@ -94,6 +102,37 @@ describe(_.startCase(filename), function () {
...
+
+    it('if subsite, bootstrapPath runs twice', function () {
+      siteService.getParentSite.returns(_.cloneDeep(sitesFake[3]));
+      sandbox.stub(lib, 'bootstrapPath').returns(Promise.resolve({}));
+
+      return fn(_.cloneDeep(sitesFake[3])).then(() => {
+        sinon.assert.calledTwice(lib.bootstrapPath);
+      });
+    });
+
```
manuelurenah: Please remove this empty line

lib/routes.js
```diff
@@ -129,8 +129,23 @@ function addCORS(site) {
...
+
+  if (parent && parent.dir) {
+    const controller = files.tryRequire(parent.dir);
+
+    // merge in subsiteController options
+    if (site.subsite) {
+      const subsiteController = files.tryRequire(site.dir);
+
+      // default _.merge overrides values in arrays :facepalm:
+      _.mergeWith(controller, subsiteController, function (par, sub) {
```
manuelurenah: Please make this an arrow function
```suggestion
      _.mergeWith(controller, subsiteController, (par, sub) => {
```

lib/routes.js
```diff
@@ -129,8 +129,23 @@ function addCORS(site) {
...
+        if (_.isArray(par) && _.isArray(sub)) {
```
manuelurenah: We could also use `Array.isArray` which is native JS instead of `lodash`, just like we do in line 151
```suggestion
        if (Array.isArray(par) && Array.isArray(sub)) {
```

lib/routes.js
```diff
@@ -129,8 +129,23 @@ function addCORS(site) {
...
+
+  if (parent && parent.dir) {
+    const controller = files.tryRequire(parent.dir);
+
+    // merge in subsiteController options
+    if (site.subsite) {
+      const subsiteController = files.tryRequire(site.dir);
+
+      // default _.merge overrides values in arrays :facepalm:
+      _.mergeWith(controller, subsiteController, (par, sub) => {
```
fxisco: should we ask if the controller is defined before this or it doesn't matter?

lib/services/sites.js
```diff
@@ -74,23 +79,68 @@ function getSites() {
...
       log('warn', `A protocol property ('http' or 'https') was not found for site '${site}', it will default to 'http'`);
       siteConfig.protocol = 'http';
     }
 
-    // add site config to the sites map
-    fullConfig[siteConfig.slug] = siteConfig;
+    return siteConfig;
+  }
+
+  files.getFolders(instanceSitesFolder).forEach(site => {
```
fxisco: **Suggestion**: run a `.map` to create an array with `subsiteDir` per each file. Then run a separate `.foreach` to check if `files.fileExists`. This way we avoid having two nested for loops.

jpope19: I'm not sure there's a good way around having O(n^2) here... I've been messing around with it for a little while. `subsiteDir` has to be looped to get all of the subsites inside of the directory, whether that happens here or in another for loop, it'll still be nested :(

lib/routes.js
jpope19: added check!

lib/routes.js
```diff
@@ -129,8 +129,25 @@ function addCORS(site) {
...
+    // merge in subsiteController options
+    if (site.subsite) {
+      const subsiteController = files.tryRequire(site.dir);
+
+      if (subsiteController) {
+        // default _.merge overrides values in arrays :facepalm:
+        _.mergeWith(controller, subsiteController, (par, sub) => {
+          if (Array.isArray(par) && Array.isArray(sub)) {
+            return par.concat(sub);
+          }
```
reubenson: needs to `return par` if sub is not an array for whatever reason?

package.json
```diff
@@ -16,7 +16,7 @@
   "author": "New York Media",
   "license": "MIT",
   "dependencies": {
-    "amphora-auth": "^1.0.0",
+    "amphora-auth": "git+ssh://git@github.com:clay/amphora-auth.git#subsite-support",
```
reubenson: will also want to make sure we bump amphora-auth before releasing this

lib/routes.js
jpope19: Should be ok as is, tested with String params during dev

https://lodash.com/docs/4.17.15#mergeWith

If customizer returns undefined, merging is handled by the method instead. The customizer is invoked with six arguments:
(objValue, srcValue, key, object, source, stack).

Note: This method mutates object.

lib/services/metadata.js
```diff
@@ -59,6 +59,7 @@ function publishPage(uri, publishMeta, user) {
 function createPage(ref, user) {
   const NOW = new Date().toISOString(),
     users = [userOrRobot(user)],
+    site = sitesService.getSiteFromPrefix(ref.substring(0, ref.indexOf('/_pages'))),
```
manuelurenah: Do you think it would be a good idea to use `getPrefix` method from `clayutils` in here?
```js
site = sitesService.getSiteFromPrefix(getPrefix(ref)),
```

lib/services/metadata.js
```diff
@@ -81,6 +83,7 @@ function createPage(ref, user) {
 function createLayout(ref, user) {
   const NOW = new Date().toISOString(),
     users = [userOrRobot(user)],
+    site = sitesService.getSiteFromPrefix(ref.substring(0, ref.indexOf('/_layouts'))),
```
manuelurenah: Same suggestion in here

package.json
jpope19: updated to ^1.2.0

PR #683, opened 2020-01-27
remove unused dependency

macgyver:
I was just chatting with Jordan about handlebars version disparities across the clay ecosystem and noticed that `amphora` specifies a handlebars dependency but never requires the library, and the tests all pass when it's removed - so that's what this PR does!

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,6 @@
     "express-flash": "0.0.2",
     "express-session": "^1.14.1",
     "glob": "^7.0.0",
-    "handlebars": "^4.0.5",
     "highland": "^2.11.1",
     "ioredis": "^4.0.0",
     "js-yaml": "^3.2.0",


jpope19:
@macgyver We've been trying to get rid of all of the handlebars versioning disparities across all of the repos. It's been a bit more difficult than expected:

I'm pretty sure there are 4-5 PRs out there to fix this across other repos but these are the two I know of off the top of my head.

https://github.com/clay/amphora-html/pull/79
https://github.com/clay/clay-kiln/pull/1495

And working to get the starter back up and running with this fixed. It will be done eventually. Not sure if this was done to limit the handlebars version used in the other dependencies?

macgyver:
word, I had a feeling it would be a huge chore. It's a little frustrating
that clay only supports one templating language, and moreso that it's a
fairly ancient one.

I support this upgrade because of the security fixes but a more
future-thinking solution would be to update everything so that Handlebars
is a peer dependency defined by the site - handlebars seems to be basically
in maintenance mode but they do periodically release security updates, a
new security vulnerability could be discovered tomorrow that will
invalidate the work you've already done.

jpope19:
I definitely endorse the extraction of any direct handlebars dependency 100%. should only be one point of entry where you pass in the templating service during amphora initiation.

PR #684, opened 2020-04-03
Add 'renderOrder' property to schema to decide in which order components render

micaww:
This allows you to specify a `renderOrder` property in the `_component` or `_componentList` property of a layout or component's schema.yml to determine the order in which components get rendered. 

All component or component-lists that don't specify this property will default to `1`.

Ex:
```yaml
col1:
   _componentList:
      renderOrder: 1
      include:
         - some-component

col2:
   _component:
      renderOrder: 2
      include:
         - some-component
```

### DIFF
--- a/lib/services/composer.js
+++ b/lib/services/composer.js
@@ -114,7 +114,7 @@ function composePage(pageData, locals) {
  * @returns {Boolean}
  */
 function filterBaseInstanceReferences(obj) {
-  return _.isString(obj[referenceProperty]) && obj[referenceProperty].indexOf('/instances/') !== -1;
+  return _.isObject(obj) && _.isString(obj[referenceProperty]) && obj[referenceProperty].indexOf('/instances/') !== -1;
 }
 
 module.exports.resolveComponentReferences = resolveComponentReferences;
--- a/lib/services/pages.js
+++ b/lib/services/pages.js
@@ -45,7 +45,7 @@ function addOp(uri, data, ops) {
  * @returns {boolean}
  */
 function isInstanceReferenceObject(data) {
-  return _.isString(data._ref) && data._ref.indexOf('/instances/') > -1;
+  return _.isObject(data) && _.isString(data._ref) && data._ref.indexOf('/instances/') > -1;
 }
 
 /**


lib/services/composer.js
```diff
@@ -5,58 +5,108 @@ const _ = require('lodash'),
...
-          };
-
-          if (error.status) {
-            logObj.status = error.status;
-          }
-
-          log('error', `${error.message}`, logObj);
-
-          return bluebird.reject(error);
+function resolveComponentReferences(uri, data, locals, filter = referenceProperty) {
```
micaww, line 24: adding `uri` to the beginning of this is definitely a breaking change if anyone is using the composer externally. i added it to the front here because that seems to be the norm across the repo. but i can move it to the end and make it optional to clear the breaking change if that's preferred

lib/services/composer.js
```diff
@@ -71,6 +120,8 @@ function filterBaseInstanceReferences(obj) {
 module.exports.resolveComponentReferences = resolveComponentReferences;
 module.exports.composePage = composePage;
 module.exports.filterBaseInstanceReferences = filterBaseInstanceReferences;
+module.exports.referenceProperty = referenceProperty;
```
micaww, line 123: had to export this to use it for calling `resolveComponentReferences` in order to remove the breaking change for that call signature

lib/services/references.js
```diff
@@ -39,6 +39,31 @@ function listDeepObjects(obj, filter) {
...
 }
 
+/**
+ * Search through an object and find all deep key-value pairs matching a filter.
+ * @param {object} obj
+ * @param {Function} [filter=_.identity] Optional filter
+ * @param {array} [path] Path of this value
+ * @returns {object}
+ */
+function listDeepValuesByKey(obj, filter = _.identity, path = []) {
```
olsonpm, line 49: Unfortunately the order listDeepValuesByKey results in is different from listDeepObjects due to your use of recursion (depth vs breadth first).

I tested on an article - and this matters because if you change the default order then when people update their version and rely on a certain rendering order, their code will behave differently and potentially break.

olsonpm:
Whoops, this PR also needs to add documentation for the new schema property.  And i'm not sure, but this seems like the second schema property that affects behavior outside kiln, the other being version.  I note this because the kiln docs say the schema is

> the main connective tissue between components and Kiln.

so the clay team may want to keep that in mind with docs regarding the schema file

lib/services/references.js
```diff
@@ -39,6 +39,31 @@ function listDeepObjects(obj, filter) {
   return list;
 }
 
+/**
+ * Search through an object and find all deep key-value pairs matching a filter.
+ * @param {object} obj
+ * @param {Function} [filter=_.identity] Optional filter
```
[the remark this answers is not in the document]
micaww: sure. was just keeping the same definition as the `listDeepObjects` function

lib/services/references.js
micaww: hey phil. tested on an article as well. the only thing that would be affected here are refs that are properties, which we already know order shouldn't be relied upon for those. array ordering is still maintained unless there is a ref inside of a ref (which afaik would never happen before data is populated)

olsonpm: i'll get an example to show what i mean

olsonpm: so testing against [this article](https://gist.github.com/olsonpm/ecc902d978ca5e3ef2e93e0bf85f4b92)

listDeepObjects returns with the order of [these refs](https://gist.github.com/olsonpm/5f8fcf37c6ad5cbc668714fec5647b34)
whereas listDeepValuesByKey returns with [this order](https://gist.github.com/olsonpm/4e145b14013cb50b3518af9e7212e07e)

In terms of keeping the public api of amphora stable, listDeepValuesByKey needs to preserve the ordering by default, regardless whether the order *should* be relied upon.

For instance, we were relying upon the rendering order and if that changed with a minor update, that would be a frustrating bug to find.

Ultimately it's the lib author's definition of "the api" and what constitutes semver changes so if you disagree then we can let them decide.

PR #687, opened 2020-09-17
adds 'url' to log context for rendering errors

mattoberle:
When Amphora throws an error in a location where req/res context is
available we don't currently include that context.

This commit includes `url` in the log context to make it easy to
reproduce errors.


jjpaulino (approved):
🏄‍♂️

lib/responses.js
```diff
@@ -312,7 +312,7 @@ function unauthorized(res) {
  * @param {object} res
  */
 function clientError(err, res) {
-  log('error', err.message, { stack: err.stack });
+  log('error', err.message, { stack: err.stack, url: res.locals.url });
```
mattoberle, line 315: **TODO**: Is a `clientError` really worthy of generating an `error` level log on the server side or should this be a warning?

mattoberle:
> lgtm, but would also be a bit safer to use lodash `_get(res, 'locals.url')`

@reubenson yeah I was thinking about that too.
It looks like Express sets `res.locals` so it should be there: https://expressjs.com/en/api.html#res.locals

The only situation I can think of where it wouldn't would be if someone added a middleware to wipe out that attribute or set it to a non-object (which seems like a mis-use of Express and maybe _should_ result in a hard error?).

PR #689, opened 2020-12-03
Add db paginate

james-owen:
Sets us up for pagination to list endpoints like `/_components/instances`
- Calls `db.paginate()` for list requests
- Adds `prev` and `size` query params to list requests

Requires https://github.com/clay/amphora-storage-postgres/pull/64 for pagination.

### DIFF
--- a/lib/responses.js
+++ b/lib/responses.js
@@ -449,15 +449,6 @@ function listWithoutVersions() {
   return list(options);
 }
 
-function paginateComponents(previous, size) {
-  // const options = {
-  //   previous: req.query.prev, 
-  //   size: req.query.size || 0
-  // }
-
-  return list();
-}
-
 /**
  * List all things in the db that start with this prefix
  * _and_ are published
@@ -569,7 +560,6 @@ module.exports.checkArchivedToPublish = checkArchivedToPublish;
 // straight from DB
 module.exports.list = list;
 module.exports.listWithPublishedVersions = listWithPublishedVersions;
-module.exports.paginateComponents = paginateComponents;
 module.exports.listWithoutVersions = listWithoutVersions;
 module.exports.getRouteFromDB = getRouteFromDB;
 module.exports.putRouteFromDB = putRouteFromDB;
--- a/lib/responses.js
+++ b/lib/responses.js
@@ -421,7 +421,9 @@ function list(options) {
     let list,
       listOptions = _.assign({
         prefix: req.uri,
-        values: false
+        values: false,
+        previous: req.query.prev, 
+        size: req.query.size || 0
       }, options);
 
     list = db.list(listOptions);
@@ -441,10 +443,19 @@ function listWithoutVersions() {
       return [filter({wantStrings: true}, function (str) {
         return str.indexOf('@') === -1;
       })];
-    }
+    }, 
   };
 
   return list(options);
+}
+
+function paginateComponents(previous, size) {
+  // const options = {
+  //   previous: req.query.prev, 
+  //   size: req.query.size || 0
+  // }
+
+  return list();
 }
 
 /**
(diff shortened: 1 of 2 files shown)


test/fixtures/mocks/storage.js
```diff
@@ -20,6 +20,7 @@ class Storage {
     this.clearMem         = this.clear;
     this.pipeToPromise    = db.pipeToPromise;
     this.createReadStream = (ops) => this.inMem.createReadStream(ops);
+    this.paginate = (ops) => this.inMem.createReadStream(ops);
```
mattoberle, line 23: Should this be `this.inMem.paginate`?

lib/responses.js
```diff
@@ -408,6 +408,41 @@ function expectJSON(fn, res) {
...
+/**
+ * Checks the page size requested to make sure it's a Number and within the maximum.
+ * Returns the default or 0 if there's no default or size is not a Number.
+ * @param {Object} req
+ * @returns {number}
+ */
+function checkPageSize(req) {
+  const size = req.query.size || process.env.CLAY_STORAGE_PAGE_SIZE || 0;
+  const num = Math.abs(Number(size));
+  const max = process.env.CLAY_STORAGE_MAX_PAGE || null;
```
mattoberle, line 420: Right now setting `max` to `0` will result in the function returning a non-integer.

```js
const max = process.env.CLAY_STORAGE_MAX_PAGE || null;
```

^ Here `process.env.CLAY_STORAGE_MAX_PAGE_SIZE` would be equal to `"0"`, not `0`, so `typeof max === 'string'`.

```
if (max) {
```

^ This condition is `true` for `max === "0"`.

```
return num > max ? max : num;
```

^ This works because javascript will sneakily convert `"0"` to `0`.

But then we ultimately return `"0"` instead of `0` (which doesn't match the type declaration in the docstring).

mattoberle (changes requested):
Let's take some time to chat about `CLAY_STORAGE_PAGE_SIZE`, `CLAY_STORAGE_MAX_PAGE`, and `.paginate`.

I think this is really close, but we need to hammer out:

1. Compatibility with other `amphora` storage modules.
2. The best place to set / validate `MAX_PAGE_SIZE` and `DEFAULT_PAGE_SIZE`.
3. The best naming convention for those two settings.

I know you mentioned moving the settings into Amphora which I think is a decent option (if we could leave it up to the storage method to implement/use the additional `previous` and `size` opts).

lib/responses.js
james-owen: Thanks for catching this. Fixed by coercing `CLAY_STORAGE_MAX_PAGE_SIZE` to a Number first.

lib/responses.js
```diff
@@ -408,6 +408,66 @@ function expectJSON(fn, res) {
...
+      } else {
+        throw Error('Client: Invalid previous');
+      }
+    } else {
+      return;
+    }
+  } catch (error) {
+    handleError(res)(error);
+  }
+}
```
mattoberle:
```suggestion
function checkPrevious(req, res) {                                               
  const prefix = req.uri;                                                        
  const previous = req.query.prev;                                               
                                                                                 
  if (previous && prefix && !previous.includes(prefix)) {                        
    return handleError(res)(Error('Client: Invalid previous'));                  
  }                                                                              
                                                                                 
  return previous;                                                               
}
```

Do we need the `try/catch` here and the casting to `null` or would something like this work?

lib/responses.js
```diff
@@ -408,6 +408,66 @@ function expectJSON(fn, res) {
...
+
+    const max = Math.abs(Number(process.env.CLAY_STORAGE_MAX_PAGE_SIZE));
+
+    if (max) {
+      return size > max ? max : size;
+    } else return size;
+  } catch (error) {
+    handleError(res)(error);
+  }
+}
```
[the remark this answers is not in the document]
james-owen: I think relying on default/max like you propose in the first suggestion is the best way. Then we don't have to keep up with which things are invalid and why.

lib/responses.js
james-owen: Yep. This should work. Much cleaner obviously. Thanks for the suggestion.

mattoberle (approved):
:+1:

We talked about one additional test case for `CLAY_STORAGE_DEFAULT_PAGE_SIZE=undefined` and `CLAY_STORAGE_MAX_PAGE_SIZE=int` and then this seems good to go to me!

jjpaulino (approved):
🍕

PR #691, opened 2020-12-10
Add patch functionality to lists

jpope19:


lib/services/lists.js
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Appends an entry to a list and returns the updated list.
+ * @param {string} uri
+ * @param {Array<Object>|Object} data
+ * @returns {Array<Object>}
+ */
+async function addToList(uri, data) {
+  // do a db call for the list
+  const list = await db.get(uri),
+    added = list.concat(data);
```
csdswarm: It might be a good idea to make sure the item being added to the list isn't already there.

jpope19: supposedly we want to allow dupes... which makes add / delete a little difficult. decided to allow dupes to be added, and deletion will remove all duplicates.

olsonpm: yeah Chris lists aren't sets as far as clay is concerned i.e. they don't have semantics for duplicates, so dupes should be allowed

lib/services/lists.js
```diff
@@ -0,0 +1,66 @@
...
+  return list;
+}
+
+/**
+ * Add or Remove an item from a list
+ * @param {string} uri
+ * @param {Array<Object>} data
+ * @returns {Promise}
+ */
+function patchList(uri, data) {
```
olsonpm, line 43: this might be the first time I've seen PATCH implemented correctly haha

lib/services/lists.js
```diff
@@ -0,0 +1,66 @@
...
+ */
+function removeFromList(list, data) {
+  const startLength = list.length;
+
+  for (const deletion of data) {
+    list = list.filter(entry => !_isEqual(entry, deletion));
+  }
+
+  if (list.length === startLength) {
+    throw new Error('Nothing was removed from the list.');
```
olsonpm, line 31: Personally I feel throwing an error in this case will prove more trouble than it's worth.  Like if two editors end up removing the same item from the list that should work just fine rather than cause a 500.

not a strong opinion though

olsonpm (approved):

lib/services/lists.js
mrollinsiv: I would tend to agree here. If the goal is to have it removed, it was already achieved.

mrollinsiv:
Is there a PR or matching ticket to update Kiln to utilize PATCH instead of doing a PUT of the entire list?

lib/services/lists.js
mrollinsiv: How will this work with components like `tags` which has a `count`?

olsonpm: is `count` used anywhere ?  afaik it's only tracked, and I don't think it's tracked very well.  At least I remember it being inaccurate.

If we care about `count` then a generic clay list isn't a good structure for this since there's no good way to update it while taking into account concurrent modifications.  A new structure with an identifier would be necessary.

jpope19:
> Is there a PR or matching ticket to update Kiln to utilize PATCH instead of doing a PUT of the entire list?

https://github.com/clay/clay-kiln/pull/1511

james-owen (approved):

PR #693, opened 2021-02-09
Add response codes to log when 400 and 500 level errors occur

Tiffachow:


Tiffachow:
What is the reasoning for expectNoLogging assertion when there are 400 errors? @jpope19

jpope19:
> What is the reasoning for expectNoLogging assertion when there are 400 errors? @jpope19

400 errors are most likely the expected response from clay, ie 404 for page not found, and in this case they'll be logged at the network level. is there a specific instance where a log for a 400 make sense?

Tiffachow:
The expectNoLogging was applied for the 501 error as well. Is that being logged at the network level?

lib/responses.js
```diff
@@ -146,6 +146,8 @@ function notImplemented(req, res) {
   const code = 501,
     message = 'Not Implemented';
 
+  log('error', message, { code });
+
```
mattoberle, line 150: This is going to produce a log line like this:
```
{ "level": 50, "message": "Not Implemented", "code": 501 }
```

I think in general if we are in this block there was a deliberate choice _before_ this block to put us here (ie. setting `err.status = 501`).
Whatever context for _why_ we got here is only going to exist in that earlier location.

lib/responses.js
```diff
@@ -166,6 +168,9 @@ function methodNotAllowed(options) {
     } else {
       code = 405;
       message = 'Method ' + method + ' not allowed';
+
+      log('error', message, { code });
+
```
mattoberle, line 173: We've been working toward reducing the severity of client errors in the logs.
I think in this case the request logs at the CDN/proxy/load balancer will cover all the context needed, there isn't much more context available here.

lib/responses.js
```diff
@@ -189,6 +194,9 @@ function notAcceptable(options) {
     } else {
       code = 406;
       message = req.get('Accept') + ' not acceptable';
+
+      log('error', message, { code });
+
```
mattoberle, line 199: (see +171 to +173)

lib/responses.js
```diff
@@ -202,7 +210,12 @@ function notAcceptable(options) {
  */
 function denyTrailingSlashOnId(req, res, next) {
   if (_.last(req.path) === '/') {
-    sendDefaultResponseForCode(400, 'Trailing slash on RESTful id in URL is not acceptable', res);
+    const code = 400,
+      message = 'Trailing slash on RESTful id in URL is not acceptable';
+
+    log('error', message, { code });
+
+    sendDefaultResponseForCode(code, message, res);
```
mattoberle, line 218: (see +171 to +173)

lib/responses.js
```diff
@@ -218,7 +231,12 @@ function denyReferenceAtRoot(req, res, next) {
   const body = req.body;
 
   if (_.has(body, '_ref')) {
-    sendDefaultResponseForCode(400, 'Reference (_ref) at root of object is not acceptable', res);
+    const code = 400,
+      message = 'Reference (_ref) at root of object is not acceptable';
+
+    log('error', message, { code });
+
+    sendDefaultResponseForCode(code, message, res);
```
mattoberle, line 239: (see +171 to +173)

lib/responses.js
```diff
@@ -271,6 +289,8 @@ function notFound(err, res) {
   const message = 'Not Found',
     code = 404;
 
+  log('error', message, { code });
+
```
mattoberle, line 293: (see +171 to +173)

lib/responses.js
```diff
@@ -302,6 +322,8 @@ function unauthorized(res) {
     message = removePrefix(err.message, ':'),
     code = 401;
 
+  log('error', err.message, { code });
+
```
mattoberle, line 326: (see +171 to +173)

lib/responses.js
```diff
@@ -328,6 +351,7 @@ function clientError(err, res) {
 function handleError(res) {
   return function (err) {
     if (err.status && err.name !== 'NotFoundError') {
+      log('error', err.message, { code: err.status, stack: err.stack });
```
mattoberle, line 354: I'm hesitant about adding logging here because this `if` statement specifically deals with errors that have already been handled to some extent.

Assigning a `status` to an `error` is a way to explicitly tell Amphora how to communicate the error in the response. The only way to get here is by:

1. Manually constructing and throwing an `Error`.
2. `try/catch` an `Error` and setting `.status`.

We could risk duplicate logs or logging handled events at the error level by putting this line here.

mattoberle (changes requested):
@Tiffachow I think we could add the status codes to the two locations that are not commented in this PR (the `400` error and the `500` error) but the rest of the additions would be better handled by adding a request logger in your implementation's middleware stack.

lib/responses.js
Tiffachow: @mattoberle I tracked down all the places where this function was called. None of them log errors before this function is called.

lib/responses.js
Tiffachow: The only areas this function is called is in `_lists` and `_uris` logic for posting to '/' route. err.status is not manually set before this.

lib/responses.js
Tiffachow: the context for the 400 from this function and the one on line 235 are different so it would be worth logging it

lib/responses.js
Tiffachow: this i agree we can remove

lib/responses.js
Tiffachow: this i agree we can remove

lib/responses.js
Tiffachow: the context for the 400 from this function and the one on line 214 are different so it would be worth logging it

lib/responses.js
Tiffachow: this i agree we can remove

lib/responses.js
Tiffachow: this i agree we can remove

lib/responses.js
mattoberle: If I'm looking at the right places, it looks like the few spot this is called in `amphora` do not set a `.status`, meaning that they hit the `clientError` or `serverError` functions (which do log).

The block with the new `log` call is intended to be a way to force a particular response code by throwing somewhere where direct access to `res` is unavailable.

eg. Maybe in a `model.save` method one would do:

```js
if (validationFailed) {
  const err = new Error('Invalid value for X');
  err.status = 400;
  throw err;
}
```

If you find a particular location in this repository where we are setting `.status` I'm 100% open to considering logging where the error is constructed-- I just want to avoid the duplication consequences for anyone taking advantage of this as-implemented!

Tiffachow: Where would the duplication be? If anyone is setting err.status, they would also have to be manually using clay log to log the error in order for it to be duplicated right?

mattoberle: You are right, the duplication would be manual-- but opting in to log at the time of error handling gives the implementer a choice of:

1. What context to include.
2. What level to log at. 

If I log when I'm handling the error I can do something like this:

```js
log('debug', 'Invalid value for X', { x });
```

Doing that groups all like-messages under `.msg == "Invalid value for X"` and supplies the context of what `x` was. It also let's me use the `debug` log-level, because it's not something we usually need to know about.

With the logging built-in to amphora I can't choose the log level and the only way to provide context is to set the message like this:
```js
Error(`Invalid value for X: ${x}`);
```

Tiffachow:
@mattoberle I updated the tests to fit the changes but now it says it's failing at the coveralls step. Advise?

mattoberle:
I'm going to close this out because https://github.com/clay/amphora/pull/694 (which was derived from this branch) was merged.

PR #695, opened 2021-02-24
reverts breaking upgrades from v7.10.0

mattoberle:
A breaking upgrade to `handlebars` was released in error as **v7.10.0**.

The change was not intended for the public and **v7.10.0** was removed from the **latest** tag on NPM / marked as deprecated.

This commit restores the `package.json` and `package-lock.json` files that were present on the last valid release, **v7.9.0**.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "amphora",
-  "version": "7.10.0",
+  "version": "7.9.0",
   "description": "An API mixin for Express that saves, publishes and composes data with the key-value store of your choice.",
   "main": "index.js",
   "scripts": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "amphora",
-  "version": "7.10.0",
+  "version": "7.9.0",
```
james-owen, line 3: Do we want to make it version 7.10.1?

mattoberle: I think we should scrap the `7.10.x` version.
After merging this the only difference between `7.9.0` and HEAD will be updates to test files.

james-owen (approved):

PR #697, opened 2021-04-05
ON-3278 | add user info to event bus

Leland-Kwong:
* This change adds `user` data to event topics that were missing it. This makes the event bus payload more consistent and allows for granular tracking by user.

### DIFF
--- a/lib/services/db-operations.js
+++ b/lib/services/db-operations.js
@@ -105,8 +105,18 @@ function getPutOperations(putFn, uri, data, locals) {
   }
 
   // run each through the normal put, which may or may not hit custom component logic
-  return bluebird.map(cmptOrLayout, ({ key, value }) => putFn(key, value, locals))
-    .then(ops => _.filter(_.flattenDeep(ops), _.identity));
+  return bluebird
+    .map(cmptOrLayout, ({ key, value }) => putFn(key, value, locals))
+    .then(ops => _.filter(_.flattenDeep(ops), _.identity))
+    .then(ops => {
+      function addUserProp(op) {
+        op.user = locals.user;
+      }
+
+      ops.forEach(addUserProp);
+
+      return ops;
+    });
 }
 
 /**
--- a/lib/services/db-operations.js
+++ b/lib/services/db-operations.js
@@ -81,6 +81,30 @@ function splitCascadingData(uri, data) {
 }
 
 /**
+ * Adds user info onto each op object
+ *
+ * @param {Object} locals
+ * @param {Object} locals.user
+ * @returns {Function}
+ */
+function addUserInfoToOp(locals) {
+  if (!locals) {
+    return _.identity;
+  }
+
+  const { user } = locals;
+
+  function addUserProp(op) {
+    op.user = user;
+  }
+
+  return (opsList) => {
+    opsList.forEach(addUserProp);
+    return opsList;
+  };
+}
+
+/**
  * Get a list of all the put operations needed to complete a cascading PUT
  *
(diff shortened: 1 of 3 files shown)


lib/services/db-operations.js
```diff
@@ -105,8 +105,18 @@ function getPutOperations(putFn, uri, data, locals) {
   }
 
   // run each through the normal put, which may or may not hit custom component logic
-  return bluebird.map(cmptOrLayout, ({ key, value }) => putFn(key, value, locals))
-    .then(ops => _.filter(_.flattenDeep(ops), _.identity));
+  return bluebird
+    .map(cmptOrLayout, ({ key, value }) => putFn(key, value, locals))
+    .then(ops => _.filter(_.flattenDeep(ops), _.identity))
+    .then(ops => {
+      function addUserProp(op) {
```
jpope19, line 112: can you define this outside of this `getPutOperations`?

jpope19: should be able to then do:

```
  return bluebird
    .map(cmptOrLayout, ({ key, value }) => putFn(key, value, locals))
    .then(ops => _.filter(_.flattenDeep(ops), _.identity))
    .then(addUserToOps);
```

lib/services/pages.js
```diff
@@ -27,13 +27,17 @@ const _ = require('lodash'),
...
-function addOp(uri, data, ops) {
+function addOp(uri, data, ops, locals = {}) {
+  const { user } = locals;
+
   ops.push({
     type: 'put',
     key: uri,
-    value: JSON.stringify(data)
+    value: JSON.stringify(data),
+    user
```
jpope19, line 40: do we need to stringify user too?

Leland-Kwong: Good question. Do we need to have the `data` object stringified? Reason I ask is it requires the consumer to do a `JSON.parse` on the data for pretty formatting. In particular, I need to parse it for our datadog logs which means we're doing a lot of extra stringifying/parsing.

lib/services/db-operations.js
```diff
@@ -80,6 +80,23 @@ function splitCascadingData(uri, data) {
...
 
+/**
+ * Adds user info onto each op object
+ *
+ * @param {Object} locals
+ * @returns {Function}
+ */
+function addUserInfoToOp(user) {
+  function addUserProp(op) {
+    op.user = user;
```
Leland-Kwong, line 91: Just a note that we'll need to stringify this as well if we do it in the [addOp function](https://github.com/clay/amphora/pull/697/files#diff-1743826468bdf1b419b84cdd1148fa5a882add8b21ab7cb731371f66a885d2dfR39)

jpope19: we should be ok without stringify! thanks for bearing with me

lib/services/pages.js
jpope19: we do need to parse data, for it's going into postgres as jsonb. user however is not, only key/value is being used from the op to store in the db [here](https://github.com/clay/amphora-storage-postgres/blob/master/postgres/client.js#L141)

Leland-Kwong:
Thanks for the review. I'm good with the changes on my end. Happy to merge as soon as I get the green light from you guys.

jpope19:
> Thanks for the review. I'm good with the changes on my end. Happy to merge as soon as I get the green light from you guys.

@Leland-Kwong hey, any chance you can remove this line here so the circle tests pass? thanks in advance https://github.com/Leland-Kwong/amphora/blob/ON-3278/add-user-info-to-event-bus/.circleci/config.yml#L60

jpope19 (approved):

PR #701, opened 2022-06-14
add uri and url to data sent to publish hook

KerenLerner:
Add `uri` and `url` to data that is sent to the publish hook


james-owen:
What's the goal with this change?

PR #703, opened 2022-07-21
run model hooks as usual on layout-level components being saved

### LINKED ISSUE #700: "model.save" is not called when publishing a layout
## Description

When publishing a page, all components in the page are saved as new published versions and the `model.save` hook is called for each of them. The same is not true of layouts - the data from the draft versions of each component is saved as a published version without passing through `model.save`

## Steps to Reproduce

1. find a component that exists on a layout and put some innocuous logging into that component's `model.save` hook
2. Go to any page using that layout and edit the layout
3. publish the layout
4. view the logs and verify you do not see the logging you added in step 1.

## Expected Behavior

a component's `model.save` hook should be fired any time that component is saved

## Next Steps
I'm not super familar with the amphora architecture but when publishing a **page**, there is a reference to the component's PUT behavior (`components.cmptPut`) in the publishing process https://github.com/clay/amphora/blob/master/lib/services/pages.js#L157

and I don't see a similar reference in the publishing process for a **layout** https://github.com/clay/amphora/blob/master/lib/services/layouts.js#L76

I tried simply replacing `put` there with `components.cmptPut` and my use case seems to be resolved, but I'm not sure what other effects this might have - do you think that is a decent solution?

macgyver:
fixes #700

### DIFF
--- a/lib/services/layouts.js
+++ b/lib/services/layouts.js
@@ -6,8 +6,9 @@ const _ = require('lodash'),
   uid = require('../uid'),
   files = require('../files'),
   models = require('./models'),
+  components = require('./components'),
   meta = require('./metadata'),
-  { getLayoutName, replaceVersion } = require('clayutils'),
+  { getLayoutName, replaceVersion, isComponent } = require('clayutils'),
   bus = require('./bus');
 
 /**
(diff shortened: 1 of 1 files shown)


jjpaulino (approved):
LGTM

PR #713, opened 2024-08-01
🍕 Fix pre-publish site controller reference

jjpaulino:
This function is not being executed in our clay instance because it is not being picked up and assigned to the site's controller. It's a function to add metadata to a page's meta.

### DIFF
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -177,6 +177,7 @@ function addSiteController(router, site, providers) {
       _.assign(site, _.pick(controller, 'resolvePublishing'));
       _.assign(site, _.pick(controller, 'resolvePublishUrl'));
       _.assign(site, _.pick(controller, 'modifyPublishedData'));
+      _.assign(site, _.pick(controller, 'assignToMetaOnPublish'));
     }
   }
 


james-owen (approved):
Nice catch.

lib/routes.js
```diff
@@ -177,6 +177,7 @@ function addSiteController(router, site, providers) {
       _.assign(site, _.pick(controller, 'resolvePublishing'));
       _.assign(site, _.pick(controller, 'resolvePublishUrl'));
       _.assign(site, _.pick(controller, 'modifyPublishedData'));
+      _.assign(site, _.pick(controller, 'assignToMetaOnPublish'));
```
macgyver, line 180: i know this is just like.. the pattern set by whoever came before, but wouldn't this be written more simply as

```js
    _.assign(site, {
      providers,
      // things to remember from controller
      _.pick(controller, 'resolvePublishing', 'resolvePublishUrl', 'modifyPublishedData', 'assignToMetaOnPublish')
    );
```

or like.. w/o lodash...

```js
    Object.assign(site, {
      providers,
      // I don't know why it's important that we not just assign everything from the controller
      resolvePublishing: controller.resolvePublishing,
      resolvePublishUrl: controller.resolvePublishUrl,
      modifyPublishedData: controller.modifyPublishedData,
      assignToMetaOnPublish: controller.assignToMetaOnPublish
    });
```

PR #715, opened 2024-09-18
Allow config option for skipping upgrades

### LINKED ISSUE #714: Allow opting out of component upgrades via amphora settings
## Is your feature request related to a problem? Please describe.

Upgrades, while convenient, are a bit of an anti-pattern. They write back to the database on a read. We generally try to avoid this kind of behavior and would like to just upgrade our components manually using migration scripts.

## Describe the solution you'd like

Allow for a config setting that will let us opt-out of using upgrades.

elgreg:
Fixes #714 - let's a site set `disableUpgrades` to short-circuit upgrading.

### DIFF
--- a/lib/services/upgrade.js
+++ b/lib/services/upgrade.js
@@ -189,6 +189,10 @@ function upgradeData(schemaVersion, dataVersion, ref, data, locals) {
  * @return {Promise}
  */
 function checkForUpgrade(ref, data, locals) {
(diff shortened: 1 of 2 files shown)


james-owen (approved):

elgreg:
Anything needed for this to merge? Slate would like to upgrade amphora to take advantage of the no-upgrade option (ironically).

